PR #131, opened 2016-04-09
fixing node-fetch dependency upgrade

asood123:


package.json
```diff
@@ -113,6 +113,7 @@
     "mocha": "^2.4.5",
     "mocha-circleci-reporter": "0.0.1",
     "nightwatch": "^0.8.15",
+    "node-fetch": "^1.5.0",
```
sedubois: Is this required?

asood123: Come to think of it, this should work with or without that line. I'll remove it in another PR.

PR #143, opened 2016-04-14
Donation boxes now keep the same height

carlosascari:


piamancini:
I wonder if we should force the text to be in just one line... maybe cap
the # of characters? What are your thoughts on this?

P.
piamancini.com
t: @piamancini
s: piamancini

carlosascari:
Well, in case you are suggesting to keep it all in one line in order to keep the height of the donations boxes, it won't help since the tier buttons as well as the overall width of the viewpoint changes it.

I don't think a single line looks too well with the current design, that said: Definitely cap the number of characters, something that gives 2 to 3 lines is plenty enough IMO.

piamancini:
ok let's leave it like this then for now. it's good enough. we can iterate later

PR #144, opened 2016-04-14
users: default random avatars

noeldelgado:
- https://monosnap.com/file/z0p4ZNWp7eGbNeLS3nYUsjQNzVoW4Z


asood123:
It would be nice to be consistent on the avatar that we show for each user. Instead of using random, could we mod on the email address, so it'll always produce the same avatar for a user?

piamancini:
I like that idea 

P.
piamancini.com
w:+54911 5317 4620 
t: @piamancini 
s: piamancini

noeldelgado:
@asood123 ok, I just update it. 
I am using the `user.id` for now because the `email` isn't present on the `user` object right now.

frontend/src/components/UserPhoto.js
```diff
@@ -1,17 +1,20 @@
...
-    url: '/static/images/default_avatar.svg',
     addBadge: false
   };
 
   render() {
-    const { className = '', url, addBadge } = this.props;
-    const avatar = (url || '/static/images/default_avatar.svg');
-
+    const { className = '', user, addBadge } = this.props;
+    const userId = (user.id || 0).toString();
```
asood123: Is toString necessary? We could keep it an Int and simplify the library by modding id with 4.

noeldelgado: @asood123 I just simplified the function as you requested. :+1:

PR #148, opened 2016-04-15
French version for the public page

xdamman:


xdamman:
Thanks for fixing the test!

PR #149, opened 2016-04-15
reverting back to tier.name

asood123:


xdamman:
Can you explain why that change?

piamancini:
because it was pulling from tier.title and it reads "Become a Members"
instead of "Become a Member"

P.
piamancini.com
t: @piamancini
s: piamancini

piamancini:
​

P.
piamancini.com
t: @piamancini
s: piamancini

xdamman:
Then we should change the title value in the tier definition in the DB. It doesn't feel right to use the name here.

asood123:
tier.name: what goes in the header? Like "Become a backer". Title is what we refer to those group of people as: "Backers". This is how it was until yesterday when we changed it thinking it somehow got flipped in the new design.

xdamman:
This is all very confusing.

To me the name of a tier should be "backer", "sponsor" (the backer tier, the sponsor tier, always lower case).
Then there should be a label which is how it is displayed. Only the label can have space and different case. I thought that "title" was today meant for being the label. And then there is a description.

PR #164, opened 2016-04-18
Currency symbol displayed as prefix in input for Donation Picker

carlosascari:


frontend/src/css/components/DonationPicker.css
```diff
@@ -10,6 +10,10 @@
 }
 
 .DonationPicker-input.Input::before {
+<<<<<<< Updated upstream
+=======
+  content: attr(data-prefix);
+>>>>>>> Stashed changes
```
asood123: Need to remove git artifacts

frontend/src/components/DonationPicker.js
```diff
@@ -70,20 +70,26 @@ function customField({onChange, value, frequency, currency, showCurrencyPicker,
...
   }, {
     label: 'EUR',
-    value: 'EUR'
+    value: 'EUR',
+    symbol: '€'
   }, {
     label: 'MXN',
-    value: 'MXN'
+    value: 'MXN',
+    symbol: '$'
```
asood123: This is fine when we show the currency picker (it's almost never though)

We need to tie the currency symbol to the group.currency. `format_currency` is likely the best place for it.

carlosascari:
@asood123 Moved the currency data to the format_currency lib

PR #170, opened 2016-04-19
Donate with PayPal + css fix

xdamman:


frontend/src/components/Tiers.js
```diff
@@ -1,128 +0,0 @@
-import React, { Component } from 'react';
```
asood123: Did you do a search for this everywhere? Looks like we are still using `Tiers` on the DonatePage.js in containers/

PR #173, opened 2016-04-19
Fixed bug; made more fault tolerant

carlosascari:
Infinite loop bug is resolved
@ character in twitter handle is now a valid handle and is parsed
Widen dot list so as to neatly show more dots
Detects missing or 404 images and fixes self and makes sure it doesn't repeat on the same image src

### DIFF
--- a/frontend/src/components/ImagePicker.js
+++ b/frontend/src/components/ImagePicker.js
@@ -3,12 +3,13 @@ import ImageUpload from './ImageUpload';
 import _ from 'underscore';
 
 const REG_VALID_URL = /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/;
-const REG_VALID_TWITTER_USERNAME = /^[a-zA-Z0-9_]{1,15}$/;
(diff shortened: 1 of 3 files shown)


frontend/src/components/ImagePicker.js
```diff
@@ -3,12 +3,13 @@ import ImageUpload from './ImageUpload';
 import _ from 'underscore';
 
 const REG_VALID_URL = /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/;
-const REG_VALID_TWITTER_USERNAME = /^[a-zA-Z0-9_]{1,15}$/;
+const REG_VALID_TWITTER_USERNAME = /^@?([a-zA-Z0-9_]{1,15})$/;
```
asood123: This is going to break both the frontend validator and the backend User model. Backend expects the string without an @ symbol. Allowing users to enter it in the field makes sense and ideally we'd fix this both front and backend. For now, easiest fix is likely to remove the @ symbol before passing to the form. What do you think?

asood123: Given the time constraint, I'm going to merge this and come back and find a fix for it.

PR #175, opened 2016-04-19
Don't bundle unused lodash code

dasilvacontin:
Transforms the imports so that the functions are cherry picked. From [`babel-plugin-lodash`'s README](https://github.com/lodash/babel-plugin-lodash):

Transforms:

```js
import _ from 'lodash';
import { add } from 'lodash/fp';

let add1 = add(1);
_.map([1, 2, 3], add1);
```

roughly to

```js
import _add from 'lodash/fp/add';
import _map from 'lodash/map';

let add1 = _add(1);
_map([1, 2, 3], add1);

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,7 @@
 {
   "presets": ["es2015", "stage-0", "react"],
   "plugins": [
-    "add-module-exports"
+    "add-module-exports",
+    "lodash"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "babel": "^5.8.23",
     "babel-eslint": "^4.1.5",
     "babel-plugin-add-module-exports": "^0.1.2",
+    "babel-plugin-lodash": "^2.3.0",
     "chai": "^3.3.0",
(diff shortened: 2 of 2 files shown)


asood123:
@dasilvacontin Is this different than https://github.com/OpenCollective/website/blob/master/frontend/src/components/ImageUpload.js#L3?

dasilvacontin:
Sorry, I gave a quick look to https://github.com/OpenCollective/website/blob/a45f891fc0a81bae1585e0b6b007460a328208c0/server/utils/api_url.js#L1, didn't notice it was a server file, and assumed all imports were done like that, yay.

Still, with this you could forget about manually writing the import paths, since the plugin will do that for you, and just `import _ from 'lodash'`.

xdamman:
Good reminder and great to add that to Babel. Thanks David!

arnaudbenard:
I have noticed that `underscore` is used as well https://github.com/OpenCollective/website/blob/5583e347eab3ffb1e0f4dce453c5893c5debe849/frontend/src/components/ImagePicker.js

PR #177, opened 2016-04-19
fix paypal button

asood123:


asood123:
Merging this without completing tests, since we need to get that button working.

PR #178, opened 2016-04-19
logic would not render button when group had a stripe and a paypal

carlosascari:


frontend/src/components/public_group/AmountPicker.js
```diff
@@ -44,7 +44,7 @@ export default class AmountPicker extends React.Component {
         )}
 
         <div className='max-width-1 mx-auto' style={{position: 'absolute', left: '0', right: '0', margin: '-25px auto'}}>
-          {!hasStripe && (
+          {(!hasStripe || (hasPaypal && hasStripe)) && (
```
asood123: Isn't that same thing as `!hasStripe && hasPaypal`?

carlosascari: No, not if both hasStripe and hasPaypal are true, I believe this is the case during the tests, the reason the button is missing.

xdamman: if both `hasStripe` and `hasPaypal` are true, then it shouldn't show the `Pay With PayPal` button.

sedubois:
I fixed the e2e testing scripts on API master, now circleci properly shows that there was a failing e2e test.

frontend/src/components/public_group/AmountPicker.js
asood123: In the current setup, both `hasStripe` and `hasPaypal` should never be true at the same time.

`hasPaypal` requires a different authentication than what a host gives us normally to pay out expenses. This is because Paypal has different APIs (and authentication) for receiving payments and paying out money. When we setup Paypal as a receiving mechanism, we store it in `ConnectedAccounts` table. When we setup Paypal to payout expenses, we store it in the `PaymentMethods` table.

It's going to be confusing for our users because they'll end up auth'ing paypal twice. I hope there's a way to do this in one go but I haven't had a chance to look at it yet.

@arnaudbenard please correct if I'm mistaken about any of this.

xdamman: In the context of the website, hasPaypal should refer to whether or not this collective accepts donations via PayPal. 

What you are talking about should only apply to the app. That's the only place where there is the concept of paying out using PayPal.

asood123: That's right. That's why both `hasPaypal` and `hasStripe` should never be true at the same time *in our current setup*. We'll update the logic when we accept both for a collective.

asood123:
I'm going to close this PR. Here's a new PR that fixes the logic and the tests: https://github.com/OpenCollective/website/pull/179

PR #181, opened 2016-04-22
Fix event tracking and logo for sponsors

xdamman:
Recording a page view as per 

And making the logos of the sponsors rectangular instead of rounded:


xdamman:
Fixed in https://github.com/OpenCollective/website/pull/194

PR #183, opened 2016-04-26
Github banner: wrapping avatars and logo in a svg…

xdamman:
…to make sure we can return a retina version of the image without having to specify dimensions in the markdown

Fixes https://github.com/OpenCollective/OpenCollective/issues/44


server/controllers/banner.js
```diff
@@ -46,25 +47,55 @@ module.exports = {
...
+          imageHeight = 64;
+          imageWidth = Math.round(dimensions.width / dimensions.height * imageHeight);
+        } 
+        else {
+          imageHeight = 64;
+          imageWidth = 64;
+        }
+
+        const base64data = new Buffer(data).toString('base64');
+        const svg = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="${imageWidth}" height="64">
```
dasilvacontin: Why hardcode the height to `64`? Shouldn't it be `imageHeight`?

server/controllers/banner.js
```diff
@@ -46,25 +47,55 @@ module.exports = {
...
+        var imageWidth, imageHeight;
+        if(tier === 'sponsor') {
+          const dimensions = sizeOf(data);
+          imageHeight = 64;
+          imageWidth = Math.round(dimensions.width / dimensions.height * imageHeight);
+        } 
+        else {
+          imageHeight = 64;
+          imageWidth = 64;
+        }
```
dasilvacontin: Block would be simpler as:
```js
let imageWidth = 64;
let imageHeight = 64;
if (tier === 'sponsor') {
  const dimensions = sizeof(data);
  imageWidth = Math.round(dimensions.width / dimensions.height * imageHeight);
}
```

server/routes.js
```diff
@@ -54,7 +54,7 @@ module.exports = (app) => {
   app.get('/:slug/:tier/banner.md', mw.cache(300), mw.fetchGroupBySlug, mw.fetchUsers(), banner.markdown);
   app.get('/:slug/:tier/badge.svg', mw.cache(300), mw.fetchUsers(), banner.badge);
   app.get('/:slug/badge/:tier.svg', mw.cache(300), mw.fetchUsers(), banner.badge);
-  app.get('/:slug/:tier/:position/avatar(.png)?(.jpg)?', mw.cache(300), mw.ga, mw.fetchUsers({cache: 300}), banner.avatar);
+  app.get('/:slug/:tier/:position/avatar(.:format)?', mw.cache(300), mw.ga, mw.fetchUsers({cache: 300}), banner.avatar);
```
sedubois: Does this yield the intended behaviour if e.g calling `GET /:slug/:tier/:position/avatar.blah`? Route `.../avatar\.(png|jpg|svg)` [should work](http://forbeslindesay.github.io/express-route-tester/). And according to [the doc](http://expressjs.com/en/api.html#req.params) format can be fetched with something like `req.params[0]` if you create a capture group (I guess with parentheses like `.../avatar\.\((png|jpg|svg)\)`).

server/controllers/banner.js
xdamman: much better indeed, thanks!

server/routes.js
xdamman: You are right. But `req.params[0]` is not great to read in the code. Here is a better way to do it:

    app.get('/:slug/:tier/:position/avatar(.:format(png|jpg|svg))?', (req, res) => {
      console.log(req.params.format);
    })

PR #184, opened 2016-04-26
Custom donation distribution

carlosascari:
Uncomment lines in AmountPicker to see multiple collectives, since only the currently viewed collective is displayed, its slider is disabled.


frontend/src/components/Slider.js
```diff
@@ -0,0 +1,101 @@
+import React, { Component, PropTypes } from 'react';
+
+export default class Slider extends Component {
```
sedubois: I'd use stateless functional components:
- cleaner syntax 
- enforces not storing state within component
- will be subject to performance optimisation in the future
https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components

frontend/src/components/public_group/AmountPicker.js
```diff
@@ -42,45 +41,15 @@ export default class AmountPicker extends React.Component {
...
-            </StripeCheckout>
-          )}
-
-          <div className='PublicGroupForm-disclaimer h6 mt2' style={{minHeight: 38}}>
-            {i18n.getString('disclaimer')} {group.host.name} {stripeDescription} {i18n.getString('for')} {group.name}. {cancellationDisclaimer}
-          </div>
-        </div>
+        <DonationDistributor 
+          amount={amount}
+          method={(hasPaypal) ? 'paypal' : (hasStripe) ? 'stripe' : 'paypal'}
```
sedubois: why does it show paypal even if `hasPaypal` is falsy?

frontend/src/components/public_group/DonationDistributor.js
```diff
@@ -0,0 +1,447 @@
...
+    const comissionAmount = this.getCommissionAmount();
+    const formattedComissionAmount = formatCurrency(comissionAmount, currency, {compact: true});
+    return (
+      <div>
+        <div className='DonationDistributor-method-label'>Commission</div>
+        <div className='DonationDistributorItem-container' style={{marginTop: 0}}>
+          <div className='flex'>
+            <div className='flex-auto'>
+              <div className='flex flex-column'>
+                <div className='flex'>
```
sedubois: I don't know much about HTML/CSS markup, but why are so many levels necessary?

frontend/src/components/public_group/AmountPicker.js
carlosascari: If the collective doesn't have paypal it tests if the collective has stripe enabled otherwise it fallsback to paypal

frontend/src/components/public_group/DonationDistributor.js
carlosascari: Yes, to get the right layout, since it is using flex boxes.
See http://www.basscss.com/#basscss-flexbox for examples

frontend/src/components/public_group/AmountPicker.js
sedubois: Why in that case display a button that will yield an error anyway?

carlosascari: It wont yield an error: https://github.com/carlosascari/opencollective-website/blob/custom-donation-distribution/frontend/src/components/public_group/DonationDistributor.js#L32

It should disable itself, if neither stripe or paypal is setup.

It is required to use one of the two payment methods: https://github.com/carlosascari/opencollective-website/blob/custom-donation-distribution/frontend/src/components/public_group/DonationDistributor.js#L16

xdamman:
How can we test this?

Can we make a test with `/wwcode` and all its chapters? (see https://opencollective.com/wwcode)

carlosascari:
@xdamman You can test it by clicking on the donate button you would use to make a donation for either paypal or stripe enabled accounts. It is setup so only the main collective is passed, so the sliders are disabled

I left a comment you can uncomment so you can see the multiple sliders in action. AmountPicker has the component with its required properties set:
https://github.com/OpenCollective/website/pull/184

xdamman:
Can you make a separate boolean for enabling sliders?

Also, how can we define the default values for each collective? E.g.
collective1: 10%
collective2: 20%
...

Depending on the value of `showSliders`, the user should be able to edit those default values or not.

carlosascari:
The property `editable` will enable sliders when it is true, however if there is only a single collective, it will disable itself, since it makes no sense to allow you to distribute something amongst one group.

Defining individual values for each collective does not exist, they are evenly distributed:
https://github.com/carlosascari/opencollective-website/blob/custom-donation-distribution/frontend/src/components/public_group/DonationDistributor.js#L359

I can quickly modify the line #359
`value: 1 / collectives.length`

To:
`value: !isNaN(collective.value) ? collective.value : 1 / collectives.length`

Where `collective.value` is a number between 0 and 1 representing its percentage.

This way you can specify the beginning amount of distribution, however anyone using the component MUST make sure the values sum up to 1.

Would you like me to make this change?

xdamman:
Yes please. Thanks!

carlosascari:
@xdamman `/wwcode` can be tested, however we'd need a way to receive all the chapters, or simply:
https://github.com/carlosascari/opencollective-website/blob/custom-donation-distribution/frontend/src/components/public_group/AmountPicker.js#L22

Add them to the collectives array, thats all. You will be able to play with the distribution of the donation. Just keep in mind, there is, as far as I am aware, nothing that handles a distributed donation.The donation will be made to the main collective, in this case `/wwcode`.

I guess, making the donation to the main collective and then logging the distribution amongs its children/chapters is the simplest way to get this working. 

What do you think?

xdamman:
Yes, good idea. We need to log somewhere the distribution. Maybe simply as a comment to the donation? Can you do a PR to the API for that?

carlosascari:
Sure, however I'd like to know how the component will receive the array of linked collectives, i'm guessing this is still in the drawing board?

I'll work on the PR

BTW, CircleCI failed at `npm install` again

carlosascari:
@xdamman I've made the changes here and in the api repo:
https://github.com/OpenCollective/api/pull/305

A donation distribution info will be stored as JSON in the comments column

sedubois:
@carlosascari Please rebase on master, this was merged: https://github.com/OpenCollective/website/pull/197

xdamman:
Still a few bugs. @carlosascari have you tested this with a transaction in test mode? After a transaction failed (see the credit card numbers you can use to test here: https://stripe.com/docs/testing), the modal is still showing up and we cannot see the error notification.

Also design wise:
- [ ] it's weird that the $100 is not right aligned with the break down
- [ ] the "Wish to remove" shouldn't be there by default. Also instead of such prompt, when it's removable, I'd simply show the dollar amount in an editable input text
- [ ] the title of the modal should be different if the user can't decide how to distribute the money
- [ ] the disclaimer at the bottom should give the total amount without the credit card processing fees, so in this case $105 (+ $3.20 credit card processing fees)

carlosascari:
My bad, I never tested a failed transaction, I will work on this & create a new PR.

BTW `Wish To Remove`, is not default:
https://github.com/carlosascari/opencollective-website/blob/37b4ee13c38eb8b3fbd1385b4146beafb81cb691/frontend/src/components/public_group/DonationDistributor.js#L35

You have to click `Wish to Add`, to change it. That being said, adding a slider and an input box to remove or reduce the amount of the donation is not simpler IMO. Can you confirm that you'd still like to see a slider and input to reduce the commission, after enabling it?

xdamman:
"Wish to add" is a terrible UX. By default, it should just show the 5% platform fees that cannot be edited.

haven't you developed the slider anyway for the other sub collectives?

carlosascari:
The slider already exists, implementing it wasn't my concern I just wanted to be clear:

- I will make the 5% platform fee active and not editable by default.
- If the 5% fee is set to be editable, then by default it will be active and a slider as well as a input box will allow the user to modify it.
- the "Wish to add/remove" toggle will be removed in place for the slider and input box.

All the other design tasks you mentioned are done, this is the only one I wanted to clear up with you

xdamman:
Yes exactly 👍

carlosascari:
Where should I store the commission amount? 
Since stripe uses the Donations table it does not have a comments field unlike Transactions where (in the comments column) the distribution information is saved.

carlosascari:
@xdamman PR ready for review https://github.com/OpenCollective/website/pull/209

PR #186, opened 2016-04-26
fixes for code review

xdamman:
Thanks @dasilvacontin @sedubois for the code review, here are the fixes :-)

### DIFF
--- a/server/controllers/banner.js
+++ b/server/controllers/banner.js
@@ -76,15 +76,11 @@ module.exports = {
       request({url: imageUrl, encoding: null}, (e, r, data) => {
         const contentType = r.headers['content-type'];
 
-        var imageWidth, imageHeight;
-        if(tier === 'sponsor') {
+        let imageWidth = 64;
+        let imageHeight = 64;
+        if (tier === 'sponsor') {
           const dimensions = sizeOf(data);
-          imageHeight = 64;
           imageWidth = Math.round(dimensions.width / dimensions.height * imageHeight);
-        } 
-        else {
-          imageHeight = 64;
-          imageWidth = 64;
         }
 
         const base64data = new Buffer(data).toString('base64');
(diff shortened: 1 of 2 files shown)


dasilvacontin:
What about the fixed `imageHeight` I mentioned?

xdamman:
For now we want to hard code that value. All avatars and logos should not be taller than 64px. We may change that in the future.

Or did you just mean to have one global constant?

dasilvacontin:
> const svg = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="${imageWidth}" height="64">
  <image width="${imageWidth}" height="${imageHeight}" xlink:href="data:${contentType};base64,${base64data}"/>
</svg>`

I'm pointing out that the svg size is the same as the contained image size (at least the element size), and you already have `imageWidth` and `imageHeight` to hold those values (`imageHeight` is always `64` currently), so it would make sense to use `imageHeight` also for the svg element's size, and not only for the contained image size.

And when you want to change the height, with another value or with the result of a function, you'd only have to change it in its assignment to `imageHeight`, instead of also having to change the hardcoded value in the svg string. 🚢

xdamman:
Ah of course, good catch!
Fixing that.

server/controllers/banner.js
```diff
@@ -76,19 +76,15 @@ module.exports = {
       request({url: imageUrl, encoding: null}, (e, r, data) => {
         const contentType = r.headers['content-type'];
 
-        var imageWidth, imageHeight;
-        if(tier === 'sponsor') {
+        const imageHeight = 64;
```
dasilvacontin: Nice touch with the `const`!

dasilvacontin:
LGMT~

server/controllers/banner.js
```diff
@@ -76,19 +76,15 @@ module.exports = {
...
+        let imageWidth = 64;
```
asood123: Why `let` vs `var`? I thought we were going to stay consistent with `var`.

server/controllers/banner.js
xdamman: `eslint` FTW!
eslint

PR #190, opened 2016-04-26
Vauge css selector made more specific so only targeted element is modified

carlosascari:


xdamman:
Can you remove the `test-result.xml` file (and add it to `.gitignore`)?

carlosascari:
I just read your comment, I'll add it to .gitignore now

.gitignore
```diff
@@ -42,6 +42,7 @@ server/views/partials/sprite.symbol.html
 
 # test artifacts
 output/
+test-result.xml
```
xdamman: sorry it's actually `test-results.xml` in plural. Can you change?

PR #194, opened 2016-04-27
value needs to be an int

xdamman:


xdamman:
This is quite urgent so I'm merging it since tests pass and it's a simple change.

PR #191, opened 2016-04-26
Displays backer and sponsor count

carlosascari:
When a yearly income variable exists, it will also show the amount donated per year

PR #195, opened 2016-04-28
Backers count

xdamman:
Original PR https://github.com/OpenCollective/website/pull/191

### DIFF
--- a/frontend/src/components/public_group/PublicGroupHero.js
+++ b/frontend/src/components/public_group/PublicGroupHero.js
@@ -11,13 +11,14 @@ export default class PublicGroupHero extends React.Component {
     if (!group.backers.length) return;
     const backerCount = filterCollection(group.backers, {tier: 'backer'}).length;
     const sponsorCount = filterCollection(group.backers, {tier: 'sponsor'}).length;
-    const yearlyIncome = 0; // TODO pass the yearly income provided by backers, will display amount counter.
+    const yearlyIncome = group.yearlyIncome / 100;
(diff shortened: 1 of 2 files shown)


xdamman:
This requires the api to return `group.yearlyIncome`, see https://github.com/OpenCollective/api/pull/306

frontend/src/components/public_group/PublicGroupHero.js
```diff
@@ -1,7 +1,35 @@
 import React from 'react';
 import PublicTopBarV2 from '../../containers/PublicTopBarV2';
+import formatCurrency from '../../lib/format_currency';
+import filterCollection from '../../lib/filter_collection';
 
 export default class PublicGroupHero extends React.Component {
```
sedubois: would be better to have this as functional component `export default ({group}) => {...}`

xdamman: Why? What the difference?

sedubois: It enforces not storing any state in React but rather in Redux, and will be subject to performance improvements in future react versions as the functional component can be flagged "pure". There is also probably less overhead to begin with, as there is no lifecycle management. But it's still an early feature. 

https://facebook.github.io/react/blog/2015/09/10/react-v0.14-rc1.html#stateless-function-components

> This pattern is designed to encourage the creation of these simple components that should comprise large portions of your apps. In the future, we’ll also be able to make performance optimizations specific to these components by avoiding unnecessary checks and memory allocations.

PR #202, opened 2016-05-03
Revert "Custom donation distribution"

xdamman:
Reverts OpenCollective/website#184


piamancini:
IMO in the cases in which removing the fee is an option, we should also
offer the option of giving more than 5%. I should go both ways.

P.
piamancini.com
t: @piamancini
s: piamancini

PR #205, opened 2016-05-03
Uses group.tiers to determine types of backers available

carlosascari:
They are sorted ASC by tier.ranges[0]
0 backers of any tier, removes them from being displayed

Fixed bug where if currency had a space, it was treated as a digit by isNaN, using a safer regex test now. https://github.com/OpenCollective/OpenCollective/issues/65

### DIFF
--- a/frontend/src/components/public_group/PublicGroupHero.js
+++ b/frontend/src/components/public_group/PublicGroupHero.js
@@ -8,21 +8,27 @@ export default class PublicGroupHero extends React.Component {
   renderHeroStatistics()
   {
     const { group } = this.props;
-    if (!group.backers.length) return;
-    const backerCount = filterCollection(group.backers, {tier: 'backer'}).length;
-    const sponsorCount = filterCollection(group.backers, {tier: 'sponsor'}).length;
     const yearlyIncome = group.yearlyIncome / 100;
     const formattedYearlyIncome = yearlyIncome && formatCurrency(yearlyIncome, group.currency, { compact: true, precision: 0 });
-    return (
+
+    const tierCountString = group.tiers.slice()
+    .sort((tierA, tierB) => tierA.range[0] - tierB.range[0])
+    .map(tier => {
+      const count = filterCollection(group.backers, {tier: tier.name}).length;
+      return (count) ? `${count} ${count === 1 ? tier.name : `${tier.name}s`}` : '';
+    })
+    .filter(x => x)
+    .join(', ');
+
+    return (tierCountString &&
       <div className='PublicGroupHero-backer-statistics'>
         <div className='PublicGroupHero-backer-count-text'>
-          We have <b>{backerCount}</b> backer{backerCount === 1 ? '': 's'}
-          {sponsorCount > 0 && (<span> and <b>{sponsorCount}</b> sponsor{(sponsorCount === 1) ? '': 's'}</span>) }
+          We have {tierCountString}
           {yearlyIncome > 0 && ' that provide us with a yearly budget of'}
         </div>
         {yearlyIncome > 0 && (
             <div className='PublicGroupHero-backer-yearly-budget'>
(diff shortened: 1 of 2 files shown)


xdamman:
Can you order by number of members of each tier DESC, and also use a "and" to join the latest?
e.g. "10 members, 3 silver sponsors and 2 gold sponsors"

Also, the "and" should be part of the i18n (and in `strings.json`)

carlosascari:
Will do. 

Also, an Inflector should be used here, I'm just adding a letter `s` to the tier names when there is more than one, what do you think? (I couldn't find a good-enough inflector, maybe you know of one)

frontend/src/ui/strings.json
```diff
@@ -107,6 +110,9 @@
     "becomeMemberTitle": "Pourquoi devenir un supporter?",
     "becomeMemberText": "Avec votre contribution, vous nous aidez à couvrir nos dépenses. Tous les fonds sont gérés de manière responsable et en toute transparence. Tout le monde peut voir comment on dépense l'argent récolté.",
     "howWeSpend": "Voir comment on utilise l'argent",
-    "donateWithPayPal": "Payer avec PayPal"
+    "donateWithPayPal": "Payer avec PayPal",
+    "and": "et",
+    "weHave": "Nous avons",
+    "thatProvideYearlyBudget": "qui nous fournissent avec un budget annuel de"
```
xdamman: ah ah nice try :-D

(almost!)
Should be "qui nous fournissent un budget annuel de"

xdamman:
Yeah we should use a proper i18n library. There is already an issue for that but we haven't taken the time to prioritize it yet. https://github.com/OpenCollective/OpenCollective/issues/7

frontend/src/components/public_group/PublicGroupHero.js
```diff
@@ -7,22 +7,45 @@ export default class PublicGroupHero extends React.Component {
...
+      return (count) ? `${count} ${count === 1 ? tier.name : `${tier.name}s`}` : '';
+    })
+    .filter(x => x);
+
+    if (tierCountStringArray.length > 1)
+    {
+      if (tierCountStringArray.length > 2)
+      {
+        const lastCountString = tierCountStringArray.pop();
+        tierCountString = `${tierCountStringArray.join(', ')} ${i18n.getString('and')} ${lastCountString}`
```
sedubois: Note that I don't think this approach scales very well. You're lucky French is rather similar to English in grammar structure, but this won't work if word order is any different. I'd have the i18n templates be whole sentences, not individual words.

sedubois:
"Fixed bug where if currency had a space" : is it normal for currencies to have space?

carlosascari:
@sedubois I'm glad you brought it up, I'm not familiar with the EUR currency format, I though it was flipped to the USD: a comma to seperate cents and periods to group 3 digits. 

The lib Numbro, used in format_currency.js here:
https://github.com/OpenCollective/website/blob/master/frontend/src/lib/format_currency.js#L27

When currency is EUR the symbol used to group 3 digits is an empty space.

xdamman:
Don't worry too much about this. I'll merge it because I really want to fix this bug on https://opencollective.com/ConsciousnessHackingSF

frontend/src/components/public_group/PublicGroupHero.js
carlosascari: I agree 100%, however I don't see how making them sentences will help. The sentence here is dynamic, it could be one tier, or more (which will require the 'and'), all this after the "we have" words. 

The i18n sentence would have to be itself a rendering template string for something like `util.format()`, where you add parameters and depending on the languages own grammatical rules it places the data (a number in this case) in the right place.

xdamman: We should use an i18n library, see OpenCollective/OpenCollective#7

PR #207, opened 2016-05-04
Don't install selenium server in prod

sedubois:


xdamman:
Instead of doing this, wouldn't it be better to just move this script to "pretest" instead of "postinstall"?

scripts/postinstall.sh
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env bash
 
+# interrupt on any error
 set -e
 
 npm run build
 
 if [ "$NODE_ENV" = "circleci" ]; then
   # pre-install API to attempt to get its node_modules cached by circleci
   npm run test:e2e:exec api:install
```
xdamman: This is confusing.

Technically, on CircleCI, before installing `website`, you should install `api` (so it should be in `preinstall`). 
Not sure why there is a reference to `e2e` here. I'd simply invoke this with `npm run api:install` (in the same way that we have `npm run db:migrate`, or `npm run db:copy`.

scripts/pretest.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# load .env file to get NODE_ENV
+if [ -f ".env" ]; then
+  source ".env"
+fi
+
+npm run hint
+
+if [ "$NODE_ENV" = "circleci" ] || [ "$NODE_ENV" = "development" ]; then
+  npm run rexec:api install_selenium_server.sh
```
xdamman: Because we only run tests locally or on circleCI, you don't need to load the `.env` and you don't need to test for the value of `NODE_ENV`. This simplifies this script to just run one command, which could be directly put in the `package.json`.

Also, `npm run exec:api install_selenium_server.sh` is a bit misleading.
Installing the selenium server should only apply to `website` and `app` (API doesn't have a user interface that needs to be tested). This command makes it sound like you are asking `api` to install selenium.
I'd do a npm script: `npm run test:e2e:install`.

One more simplification: you can merge `npm run test:e2e` and `npm run test:e2:exec` (so that it matches `npm run test:server`)

Sorry to bother you with all this but I think that keeping things simple with clear names makes it easier to understand how the system works. It will also make it easier for new people to understand what's going on (and help avoid bugs).

xdamman:
Let's wrap this one up when you can. See also my comments on your similar PR to `app`.

asood123:
This has been sitting for a while. What do we need to do here?

sedubois:
I'll come back to it after getting promises/sequelize fixed

scripts/pretest.sh
xdamman: @sedubois can you finish this?

PR #208, opened 2016-05-05
add group.slug to PublicGroup element

phated:
As a temporary solution for the GulpJS project (until https://github.com/OpenCollective/OpenCollective/issues/63 can be implemented), @xdamman and I discussed adding the `group.slug` as a class to the hero and then adding custom CSS based on it.


phated:
Looks like the CI failed due to a timeout but I'm not sure.

phated:
Sorry, still learning the code.  I've moved the `group.slug` className higher up in the tree, to the PublicGroup container.  I've also rebased out the original change.

PR #209, opened 2016-05-05
Custom donation distribution v2

carlosascari:
Set `optionalComission` property to true to allow users to modify the commission amount.
Tested w/Stripe and paypal accounts
Using i18n strings

### DIFF
--- a/frontend/src/components/Slider.js
+++ b/frontend/src/components/Slider.js
@@ -1,0 +1,101 @@
+import React, { Component, PropTypes } from 'react';
+
+export default class Slider extends Component {
+
+  static propTypes = {
+    onChange: PropTypes.func,
+    value: PropTypes.number
+  }
+
+  static defaultProps = {
+    onChange: Function.prototype,
+    value: 0,
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {pressed: false, dragging: false};
+    this.onPress = this.onPress.bind(this);
+    this.onMove = this.onMove.bind(this);
+    this.onRelease = this.onRelease.bind(this);
+    this.onTouchPress = this.mousifyTouchEvent(this.onPress);
+    this.onTouchMove = this.mousifyTouchEvent(this.onMove);
+    this.onTouchRelease = this.mousifyTouchEvent(this.onRelease);
+  }
+
+  render() {
+    return (
+      <div 
+        ref="container"
+        className={`Slider-container ${this.props.className}`}
+        onMouseDown={this.onPress}
+        onMouseMove={this.onMove}
+        onMouseUp={this.onRelease}
+        onTouchStart={this.onTouchPress}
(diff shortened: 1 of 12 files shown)


xdamman:
- [ ] don't show the subtotal if there is only one collective
- [ ] don't show the icon on the left of "Credit card processing fee" when there is only one payment method
- [ ] big donate button's label should be "Pay ${totalAmount}", where totalAmount should include the processing fee
- [ ] Show "(2.9% + $0.30 processing fee)" next to PayPal, and next to "Credit card"
- [ ] in the disclaimer at the bottom, add within the parenthesis (+ $0.30 payment processing fees)
- [ ] the commission for OpenCollective should go from 0 to 20%
- [ ] if possible, would be good to show the $ or € symbol next to the input text when editing the amount (or don't use an input type text for consistency with the other collectives)

Also I'm not a fan of the sub headers "Commission", "Donation method". They can be visually improved. Maybe we can merge them into simply: "Fees"

xdamman:
Either you use input type text for all amounts for all collectives (with the $ sign before)
or you remove the input type text for the OpenCollective commission.

And I'd just get rid of the subheaders "Commission" and "Donation method", it's pretty explicit anyway.



(I love the sliders 👍)

Also, it would be good to have a reset button/link somewhere. Maybe the "Edit donation" link turns into "Reset default distribution" after editing the distribution? Open to other suggestions.

carlosascari:
> Show "(2.9% + $0.30 processing fee)" next to PayPal, and next to "Credit card"

I though that paying through paypal did not include a processing fee, can you confirm there will be one?

> the commission for OpenCollective should go from 0 to 20%

I assume it will default to 5% or would you like the default to be customizable via a prop?

>  if possible, would be good to show the $ or € symbol next to the input text when editing the amount (or don't use an input type text for consistency with the other collectives)

I will remove input type text and add a currency symbol as a prefix

xdamman:
- There is a paypal fee (2.9% + $0.30)
- If the OC commission is editable, the default should be 10%, otherwise it's 5% (we may change that in the future)

asood123:
Also, is this going to enable this flow for everyone? We would only want to enable it selectively.

carlosascari:
@asood123 Yes it will enable the flow for everyone, however you can specify whether or not they can modify the distribution and whether or not they can modify the commission.

I can add another boolean to not show the modal and instead render the stripe or paypal buttons, would you like me to add these changes?

I'm going to have to close this PR and resend it, this is a fork, so it will fail during CircleCI

asood123:
Yes, let's put behind a boolean. Could you create a setting for it in group.settings? (https://github.com/OpenCollective/api/pull/334).

carlosascari:
Yes, I will close this PR, create a new PR add the boolean to group.settings. I'm going to set the default to simply render the paypal and stripe buttons

PR #211, opened 2016-05-06
using French locale in moment

xdamman:


frontend/src/containers/DonatePage.js
```diff
@@ -238,6 +231,6 @@ function mapStateToProps({
     saveInProgress: users.updateInProgress,
     isAuthenticated: session.isAuthenticated,
     newUser,
-    i18n
+    i18n: i18n(group.settings.lang)
```
sedubois: Why not passing `lang` instead of `i18n`? Looks a bit strange to me to pass a whole library, a lib is not a prop.

xdamman: You are absolutely right. This whole thing is a hack because we need something out the door quickly. We should use a i18n library and save the `lang` as a setting of the group.

PR #217, opened 2016-05-10
Fixed height for UserCard-name and removed UserPhoto gray background

carlosascari:
When a sponsor didn't add a name `.UserCard-name` would collapse, a fixed height was used.

A gray rectangle bg from UserPhoto component was removed since it is visible even after an image is set when the UserCard is a sponsor, unlike any other tier where the image would cover the bg.


carlosascari:
Close in favor of https://github.com/OpenCollective/website/pull/220 (no fork)

PR #232, opened 2016-05-13
use WEBSITE_PORT var for custom port when starting server; closes #230

### LINKED ISSUE #230: PORT env var used in api and website
both servers will look for `process.env.PORT` when starting.  this can be problematic for obvious reasons.

boneskull:
**UPDATE**: Heroku needs `PORT`, so we'll use `WEBSITE_PORT` if present, otherwise `PORT`, otherwise the hardcoded default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "start": "NODE_CONFIG_DIR=./server/config node server",
-    "dev": "npm run build && PORT=3000 nf start -j Procfile_dev",
+    "dev": "npm run build && WEBSITE_PORT=3000 nf start -j Procfile_dev",
     "hint": "eslint '**/*.js' --quiet",
     "pretest": "npm run hint",
     "test": "npm run test:server && npm run test:unit",
--- a/server/app.js
+++ b/server/app.js
@@ -91,7 +91,7 @@ app.use((err, req, res, next) => {
  * Port config
  */
 
-app.set('port', process.env.PORT || 3000);
+app.set('port', process.env.WEBSITE_PORT || 3000);
 
 /**
  * Start server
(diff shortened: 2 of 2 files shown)


sedubois:
AFAIK `PORT` is a reserved env var provided by Heroku, we need to use that one

asood123:
Does this need to be resubmitted without a fork for tests to pass?

boneskull:
Yes, as it stands.  The problem is simply that the environment variables are not present.  

However... I'm not entirely convinced that those environment variables (or some subset of them) need to be present to run our test suite.  The server, OTOH, disagrees.  It would be worth seeing just what happens if we modify the server to allow certain variables to be undefined if run in development (or test) mode.  It's definitely useful to have the server "fail fast" when in production, but if we aren't going to actually be using an env var in test, then we can safely ignore its absence, right?

boneskull:
Anyhow, curious on your thoughts; if you'd like me to spend a little bit of time to try to get this working, I will.

cc @sedubois for input, if any

sedubois:
@boneskull hmm some tests do need env vars, e.g the Github OAuth flow e2e test uses GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET. As far as I remember here we said we have 2 choices, either allow "Permissive building of fork pull requests" in CircleCi config knowing "There are serious security concerns"according to CircleCi, or you need to re-submit a non-fork PR. @asood123 ?

asood123:
Yep, they aren't going to pass as a fork. If you duplicate these changes to a branch within our repo, they should pass.

boneskull:
oi, will do.  I've had a similar problem with Mocha; we have suite of tests running on SauceLabs, but that can't run on PRs, because security.  I ended up simply disabling the suite for PRs, because something is better than nothing.  not ideal, but I'm not sure if there is an *actual* solution to this problem out there--if there is, I'd sure like to know about it.

PR #233, opened 2016-05-13
Custom Donation Distribution

carlosascari:
skipModal property is set to true when group.settings.DonationDistributor is falsy, will only render paypal or stripe buttons
public_donation_page - removed deprecated test


carlosascari:
@asood123 the DonationDistributor component requires `group.settings.DonationDistributor`, to be true in order for it to activate, otherwise it renders the buttons just as they have been.

Im thinking of adding this to the Group modal in the api repo, so it is set as default everytime a new collective is created. It will look like this
```
    settings: {
      type: DataTypes.JSON,
+     defaultValue: JSON.stringify({
+       DonationDistributor: false
+      }),
      allowNull: true
    },
```
> server/models/Group.js

I can also set default values that are hard coded properties in this commit:

```
    settings: {
      type: DataTypes.JSON,
+     defaultValue: JSON.stringify({
+       DonationDistributor: {
+            ...
+       }
+      }),
      allowNull: true
    },
```

**editable** - Whether or not the sliders will be displayed in order to allow custom donation distribution amongst more than 1 collectives.

**optionalComission** - Whether or not the commission will be allowed to be modified.

**feesOnTop** - When `true` fees are added on top of donation, otherwise they are taken from the donation.

**collectives** - An array holding the collectives that are tied to the main collective, these are the sub collectives that are displayed when customizing the distribution of a donation. Each element is an object w/ `{name: String, avatar: String, id: Number}`

Sounds good?

asood123:
I like the simpler default of setting DonationDistributor to false. Let's document the rest of params in a wiki for now.

PR #234, opened 2016-05-16
Added middleware to catch uncaught errors

carlosascari:
Determines if the error thrown because the api server is down
Also determines if api server is up, but fetching failed
Any other error is logged

### DIFF
--- a/server/middlewares.js
+++ b/server/middlewares.js
@@ -139,14 +139,10 @@ const addTitle = (title) => {
 }
 
 const handleUncaughtError = (error, req, res, next) => {
-    if (error.name === 'FetchError')
-    {
-      if (error.code === 'ECONNREFUSED')
-      {
+    if (error.name === 'FetchError') {
+      if (error.code === 'ECONNREFUSED') {
         console.error('API Server is down')
-      }
-      else
-      {
+      } else {
         console.error('There was an error fetching from api')
       }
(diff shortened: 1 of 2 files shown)


asood123:
Thanks for the quick response. It's breaking the 404 page in tests. Adding @arnaudbenard as well

server/middlewares.js
```diff
@@ -137,6 +137,38 @@ const addTitle = (title) => {
     next();
   }
 }
+
+const handleUncaughtError = (error, req, res, next) => {
+    if (error.name === 'FetchError')
+    {
```
asood123: We usually put braces on the previous line. http://eslint.org/docs/rules/brace-style#tbs

carlosascari: fixed :+1:

carlosascari:
@asood123, @sedubois, I'm not sure what happened here

> Your build ran 168 tests in junit with 0 failures

...

> Running:  Redirects to Github after clicking connect
TEST FAILURE: 1 assertions failed, 59 passed (55.958s)
 ✖ connect_account_page
   - Redirects to Github after clicking connect
     Testing if the URL contains "http://localhost:3000/testcollective"

arnaudbenard:
When I run `website` without `api`, I get the standard error message. Looks like the proper fix 👍

PR #242, opened 2016-05-17
another fix for banner

asood123:


asood123:
merging this since it's blocking an external urgent fix.

PR #246, opened 2016-05-17
disable github nightwatch test while we are changing the flow

asood123:


asood123:
Merging this to get all the other PRs to pass.

PR #247, opened 2016-05-18
Added padding so when button label overflows the text in tier box

carlosascari:
…still shows
Added code to try and keep the Tier Button single line
Also added style updates for PublicGroup (rounded tier buttons as originally designed)

### DIFF
--- a/frontend/src/components/AsyncButton.js
+++ b/frontend/src/components/AsyncButton.js
@@ -7,7 +7,8 @@ export default ({
   customClass,
   children,
   color,
-  disabled
+  disabled,
+  style
 }) => {
   const noop = () => {};
   const btnClass = classNames({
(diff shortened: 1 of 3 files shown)


carlosascari:
@asood123 
This will fix the issue mentioned by jaredpolivka in the user-feedback channel.

Specifically, adding a padding at the bottom of the tier box, is the fix, this way the tier bottom will never cover the content inside the tier box.

I added a quick test as well that determined if the button is too large and changes the font down in order to try and keep it all in a single line.

The original design by cuiki has rounded tier buttons so I added fix here as well.

I realize you suggested a 'more' link, however it does not fix the issue because of the absolutely positioned tier button, that is what is covering the content. It has nothing to do with the min-height of the tier box

asood123:
Makes sense. Glad you took care of it.

PR #248, opened 2016-05-19
Move Passport auth to API

sedubois:


sedubois:
Needs merge of API for tests to pass

server/routes.js
```diff
@@ -35,11 +34,12 @@ module.exports = (app) => {
    */
   app.use(robots(path.join(__dirname, '../frontend/dist/robots.txt')));
 
-  /**
-   * Authentication
-   */
-  app.get('/api/auth/:service(github)/:slug', aN.authenticateService);
-  app.get('/auth/:service(github)/callback/:slug', aN.authenticateServiceCallback);
+  // For some reason during OAuth, the redirect through piping automatically changes 
+  // the body, but not the window location -> instead redirect manually
```
sedubois: @arnaudbenard any idea why this is needed and if there's a better fix?

asood123:
code looks fine. will wait for tests to pass. then, i can rebase the `github-integration` branch on top of this

server/routes.js
arnaudbenard: I've never done redirects with piping, I found a SO question about it

http://stackoverflow.com/questions/13954094/redirect-express-js-piped-req-pipe-res-properly

sedubois: Great that works, thanks for googling ;)

sedubois: Fix: https://github.com/OpenCollective/opencollective-website/pull/253

arnaudbenard: Awesome 💃

PR #249, opened 2016-05-19
remove fees on top by default

asood123:


asood123:
Merging to stop adding inaccurate fees on top.

PR #250, opened 2016-05-19
removing separate processing fee altogether

asood123:


frontend/src/components/DonationPicker.js
```diff
@@ -7,6 +7,8 @@ import Input from './Input';
 import Select from './Select';
 
 export default ({value, currency, frequency, presets, onChange, showCurrencyPicker, i18n}) => {
+
+  console.log( "REACHED inside donationPicker");
```
sedubois: remove ?

asood123: good catch. done.

PR #256, opened 2016-05-21
don't pass our token to github call

asood123:


asood123:
Merging this one line change to get github onboarding running on staging

PR #261, opened 2016-05-23
On boarding fixes

carlosascari:
Covers the task defined here https://github.com/OpenCollective/OpenCollective/issues/81#issuecomment-220786862


carlosascari:
@asood123 @xdamman  Please review

frontend/src/containers/OnBoarding.js
```diff
@@ -55,12 +54,17 @@ export class OnBoarding extends Component {
...
+      notify
     } = this.props;
 
-    if (githubUsername && this.state.step === 2) {
+    if (githubUsername && this.state.step === 1) {
       appendGithubForm({username: githubUsername, token: params.token});
-      fetchReposFromGitHub(githubUsername);
+      fetchReposFromGitHub(githubUsername)
+      .catch((error) => {
+        this.setState({step: 0});
```
asood123: Won't this create an infinite loop? When user goes back to first page, clicking on Github instantly auths them and they are back to the first screen with same notification. I think it'll be better to give them that message inside the flow. What do you think?

frontend/src/containers/OnBoarding.js
```diff
@@ -99,18 +105,22 @@ export class OnBoarding extends Component {
 
     return validateSchema(githubForm.attributes, githubSchema)
       .then(() => createGroupFromGithubRepo(payload, attr.token))
-      .then(() => this.setState({step: 5}))
+      .then(() => this.setState({step: 4}))
```
asood123: Same infinite loop comment here. We don't support back in our flow... maybe worth thinking about?

frontend/src/containers/OnBoarding.js
carlosascari: I don't understand how it will loop infinitely, there is an `if` check that tests the step is one on line 61
I've manually tested it without any infinite loops

frontend/src/containers/OnBoarding.js
carlosascari: Not sure what you mean here, a step has been removed, instead of 5 there are 4 now (3 steps + thank you)

frontend/src/containers/OnBoarding.js
asood123: The problem is that it lands you back to the first page with the notification and then you have no options. If you click on "Connect Github" again, since you are already logged in, you are back on the first page with the same message. Is there a way to avoid that?

frontend/src/actions/github/fetch_contributors.js
```diff
@@ -11,8 +11,17 @@ export default (owner, repo) => {
     dispatch(request(owner, repo));
 
     return getThirdParty(`https://api.github.com/repos/${owner}/${repo}/contributors`)
-    .then(json => dispatch(success(owner, repo, json)))
-    .catch(error => dispatch(failure(owner, repo, error)));
+    .then(json => {
+      if (json.length >= constants.MIN_CONTRIBUTORS_FOR_ONBOARDING) {
+        dispatch(success(owner, repo, json));
+      } else {
+        throw new Error(`The repository ${repo}, does not have ${constants.MIN_CONTRIBUTORS_FOR_ONBOARDING} or more contributors`);
```
asood123: no comma needed after ${repo}

frontend/src/containers/OnBoarding.js
asood123: Oops I meant to select line 120 or so. But I think that flow works fine because you can select another repo.

frontend/src/components/on_boarding/OnBoardingHero.js
```diff
@@ -1,28 +1,23 @@
...
-          We are starting to accept new open collectives. Reserve your spot today.
-        </div>
-        <div className='-button-container'>
-          <div className='OnBoardingButton' onClick={onClickStart}>Let's do this</div>
-        </div>
-        <div className="-requirements">
-          You'll need a GitHub account, a repository with over 100 stars that you own & at least 2 contributors.
-        </div>
+export default () => {
+  return (
```
sedubois: can write directly 

````
export default () => (
  <div>...</div>
);
```

frontend/src/components/on_boarding/OnBoardingStepHeading.js
```diff
@@ -2,20 +2,18 @@ import React from 'react';
 
 export default class OnBoardingStepHeading extends React.Component {
```
sedubois: can be simplified to stateless functional component

PR #263, opened 2016-05-25
Better  flow for when user does not have enough stars

carlosascari:
Added autoclose property for Notification component

For https://github.com/OpenCollective/OpenCollective/issues/81#issuecomment-221390767


frontend/src/containers/Notification.js
```diff
@@ -5,12 +5,25 @@ import resetNotifications from '../actions/notification/reset';
 import Icon from '../components/Icon';
 
 class Notification extends Component {
```
sedubois: I don't understand the distinction between the concepts of "closed" and "hide", if it's the same thing then would be better to use a single term

frontend/src/containers/Notification.js
```diff
@@ -5,12 +5,25 @@ import resetNotifications from '../actions/notification/reset';
...
+
+  constructor(props) {
+    super(props);
+    this.state = {closed: false};
```
sedubois: If notifications state is already handled by the Redux store, I didn't understand why a React state is also needed

carlosascari: It stuck around it shouldn't be there I'll remove it

PR #273, opened 2016-05-31
use WEBSITE_PORT var for custom port when starting server; closes #230

### LINKED ISSUE #230: PORT env var used in api and website

boneskull:
See #232 for original PR.

Opening this one here for CI.


asood123:
@sedubois can you pls review and merge?

PR #277, opened 2016-06-01
Added icons as labels for inputs

carlosascari:
upload image avatar is now first option unless an social avatar is fetched
fixed ImagePicker badge bug where badge images would not resize to fit

For issue https://github.com/OpenCollective/OpenCollective/issues/103


frontend/src/components/ImagePicker.js
```diff
@@ -197,35 +197,45 @@ export default class ImagePicker extends Component {
...
+    if (!this.state.isLoading) {
       this.setState({isLoading: true});
     }
 
     getSocialMediaAvatars(newUser.id, {website: website, twitterHandle: twitter, name: profileForm.name})
     .then((response) => {
       const currentOption = this.options[this.state.currentIndex];
       const results = response.json.filter((option) => this.blacklist.indexOf(option.src) === -1 );
+      const newResults = results.filter((option) => {
+        let alreadyExists = false;
```
asood123: Is there a benefit of using `let` vs `var`? If not, could you change to `var`? It's used everywhere else in the codebase.

frontend/src/components/ImagePicker.js
```diff
@@ -197,35 +197,45 @@ export default class ImagePicker extends Component {
...
+        this.options.forEach((opt) => {
+          if (opt.src === option.src) {
+            alreadyExists = true;
+          }
+        })
+        return !alreadyExists
+      });
       let isFirstTime = false;
```
asood123: same as above.

asood123:
Besides the `let` issue, everything else looks good

frontend/src/components/ImagePicker.js
carlosascari: @asood123 
I've noticed `var` used on the server side, I've seen `let` used throughout the frontend. In ES6 `let` is favoured over `var`, and they do differ http://stackoverflow.com/a/11444416

sedubois: I prefer the `let` as well, the narrower the scoping the better. (I come from the Java world...)

asood123:
`let` makes more sense. I added it to our linter wishlist https://github.com/OpenCollective/OpenCollective/issues/76

PR #282, opened 2016-06-07
Repository fetch now goes through api instead of github directly

carlosascari:
Added action to fetch github auth user information
In order to preserve api calls, a repo a user does not own, after selecting it, if the top 3 users does not include the auth user then it is removed, otherwise api calls would have to be done in advance.

Covers the last tasks described here:
https://github.com/OpenCollective/OpenCollective/issues/81#issuecomment-221390767

And this issue:
https://github.com/OpenCollective/OpenCollective/issues/117#event-683413872

This PR is related the branch with the same name `fixes-for-github-flow` in website repo, and should be merged first, as the other changes rely on this repo, and not vice versa.

This PR relies on changes found https://github.com/OpenCollective/opencollective-api/pull/416, in order to test it you'll have to use the `fixes-for-github-flow` branch on both the `website` and `api` repositories.


frontend/src/actions/github/fetch_repos.js
```diff
@@ -1,16 +1,15 @@
...
  * Fetch a user's repo
  */
 
-export default (username) => {
+export default (username, token) => {
   return dispatch => {
     dispatch(request(username));
-
-    return getThirdParty(`https://api.github.com/users/${username}/repos?per_page=100`, {params: {type: 'owner,member'}})
+    return get(`/../github/repositories/${token}`)
```
asood123: Is there a way to avoid using `/../`? Do we even need a `website` route for this? We could straight to API.

carlosascari: Ok, I'll remove the route

carlosascari: @asood123 ok so the `api.get` method prepends whatever is in `API_ROOT` env variable. 
API_ROOT gives me localhost:3000/api not the api server but the website server and with `api/` as the parent resource.

I will create another `get` method so I can make requests for the api (localhost:/3060)

asood123: That should work without needing a new method (since all the other calls to the API use the same function). 

For ex: https://github.com/OpenCollective/opencollective-website/blob/master/frontend/src/actions/groups/get_leaderboard.js

carlosascari: @asood123 You are correct I didnt realize routes with `/api/` get piped to the api repo, please review the fix I've commited

asood123:
This looks good to be merged. Let's get the `api` PR merged before merging this one or we'll be blocked from pushing to production.

It would also be nice to have a e2e test for the github flow. Right now we only test logging into Github. Is there a way to set that up with our test account (which won't have any repos over 100 stars)?

carlosascari:
@asood123 We could set the maximum stars and contributors to zero during testing with the `NODE_ENV` variable. All that is needed then is to use a github account with at least one repository. This is enough to go through the entire github flow during testing.

asood123:
Good idea. Let's set stars to 0. We can add fake contributors to the test account to get it closer to the production environment.

PR #288, opened 2016-06-09
Revert "Pending collective"

asood123:
Reverts OpenCollective/opencollective-website#286


asood123:
Merging this to undo the last change. tests are fine - failing on paypal because we hit the sandbox too often in the last hour.

PR #292, opened 2016-06-14
Contributors are listed when group.data.githubUsers is an object:

carlosascari:
with a username as key and its value the total number of commits.
Modified UserPhoto so it listens to hover events.

For https://github.com/OpenCollective/OpenCollective/issues/122

As long as `group.data.githubUsers` is falsy it will not display. 
Contributors are sorted so core contributors are first followed by number of commits after.
Only commits are displayed, if additions and/or deletions are included they will be displayed as well.

@asood123 @boneskull  
`group.data` does not tell us whether a user is a core contributor or not. How can we determine core contributors?

### DIFF
--- a/frontend/src/components/UserPhoto.js
+++ b/frontend/src/components/UserPhoto.js
@@ -30,14 +30,14 @@ export default class UserPhoto extends React.Component {
   }
 
   render() {
-    const { className, user, addBadge } = this.props;
+    const { className, user, addBadge, onMouseEnter, onMouseLeave } = this.props;
     const avatar = this.state.avatar;
     const styles = {
       backgroundImage: `url(${avatar})`
     };
 
     return (
-      <div className={`UserPhoto bg-no-repeat bg-center relative ${user.tier} ${className} ${avatar ? 'UserPhoto--loaded' : ''} `}>
+      <div className={`UserPhoto bg-no-repeat bg-center relative ${user.tier} ${className} ${avatar ? 'UserPhoto--loaded' : ''} `} onMouseEnter={ onMouseEnter } onMouseLeave={ onMouseLeave }>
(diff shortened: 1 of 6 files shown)


asood123:
What would the definition of a core contributor be? @boneskull

xdamman:
I'd say that core contributors should have an open collective account and be MEMBERS of the collective. @carlosascari can you make it that way and make sure there is no duplicate?

asood123:
Are you still working on this, @carlosascari or should I merge?

carlosascari:
@asood123 
The task based on the mockup by Erick shows contributors ordered by importance where core contributors are listed first, then by total contributions. There is no `core` information in items from `group.data.githubUsers`, so contributors will never be ordered by being a core member, they will only be ordered by their total contributions.

Items in `group.data.githubUsers`, instead of being the number of contributions it could be an object with *contributions*, *deletions*, *additions* and a *core* boolean

It works, but it is not complete

boneskull:
What's "deletions" and "additions"?

The "core" boolean seems unnecessary if we have a definition of who is a "core" contributor elsewhere in the DB.  We can then store an *email address* (in addition to the GitHub username) in order to perform the lookup.  Granted, this should all be part of an actual schema instead of a JSON blob once things settle down.

I explained to @xdamman in Slack where to add more user data...

carlosascari:
@boneskull 
By core I was referring to the actual variable used in the code in this PR to order contributos see: https://github.com/OpenCollective/opencollective-website/pull/292/files#diff-922bbe5781cbfbe31e15ca1525fdea6fR106

Deletions and additions as well as commits are displayed on the hover card
https://github.com/OpenCollective/OpenCollective/issues/33#issuecomment-217232132

PR #296, opened 2016-06-15
change github/apply to opensource/apply

asood123:


frontend/src/routes.js
```diff
@@ -13,6 +13,9 @@ export default (
     <Route path="/subscriptions/:token" component={Subscriptions} />
     <Route path="/subscriptions" component={Subscriptions} />
     <Route path="/leaderboard" component={Leaderboard} />
+    <Route path="/opensource/apply/:token" component={OnBoarding} />
+    <Route path="/opensource/apply" component={OnBoarding} />
+    {/* Leaving github/apply routes for existing links */}
```
sedubois: can possibly also use [React redirects](https://github.com/reactjs/react-router/blob/master/docs/API.md#redirect)

PR #297, opened 2016-06-15
Profile

xdamman:
Making sure that https://OpenCollective.com/railsgirlsatl and https://OpenCollective.com/RailsGirlsAtl both work.


xdamman:
Merging this because it creates a lot of 404 on production

PR #299, opened 2016-06-16
adding redirects for consciousnesshackingSF/SV

xdamman:
When we deploy this, we need to change the slug of `consciousnesshackingsf` to `chsf` and `consciousnesshackingsv` to `chsv`


asood123:
I'm going to wait to merge this until our next push, that way we are less likely to forget to make the slug change after push.

xdamman:
Merging now and doing the change at the same time

PR #301, opened 2016-06-17
updated metadata for profile page description

xdamman:


frontend/src/containers/ProfilePage.js
```diff
@@ -39,14 +39,14 @@ export class ProfilePage extends Component {
         {belongsTo.length ? (
             <section>
               <div className="lineA">I proudly belong to these collectives...</div>
-              {belongsTo.map((collective, index) => <ProfileCard key={index} image={collective.logo} title={collective.name} subtitle={`${collective.members} Members`} footer={`${i18n.getString('since')} ${i18n.moment(collective.createdAt).format('MMMM YYYY')}`} />)}
+              {belongsTo.map((collective, index) => <ProfileCard key={index} image={collective.logo} title={collective.name} subtitle={`${collective.members} Members`} footer={`${i18n.getString('member')} ${i18n.getString('since')} ${i18n.moment(collective.createdAt).format('MMMM YYYY')}`} />)}
```
asood123: We probably want to do one string for 'member since', as translation will sometimes flip the order.

xdamman: Not in French nor Spanish.
So let's not block on that.

I did it this way because you may have something different than member (e.g. sponsor).

Also, we should use a proper i18n library to do all this.

PR #302, opened 2016-06-20
Disables create button for 5 seconds

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/140

I chose to disable the button temporarily in order to discourage multiple clicks, and not leave the user in a possible frozen state where they will have to refresh the page in order to continue.


asood123:
@carlosascari can you pls add a test for this? Either unit test or e2e is fine.

asood123:
@carlosascari going to merge this to prevent bug from happening. Can you add the test separately pls?

carlosascari:
@asood123 Will do, I'm working on the test right now

PR #307, opened 2016-06-20
Added Spanish Expense Categories

piamancini:


frontend/src/ui/expenseCategories.js
```diff
@@ -39,6 +39,24 @@ export default function(groupid) {
         'Serveur',
         'Transport'
       ];
+       break;
+    case: 114: // partidodigital
```
carlosascari: @piamancini Remove the colon `:` after the word `case`

So:
`case: 114: // partidogital` 
becomes 
`case 114: // partidogital`

piamancini: @carlosascari done, thanks!

PR #312, opened 2016-06-22
showing sponsors at different size if they are silver/gold/diamond

xdamman:


server/controllers/banner.js
```diff
@@ -82,9 +86,9 @@ module.exports = {
       request({url: imageUrl, encoding: null}, (e, r, data) => {
         const contentType = r.headers['content-type'];
 
-        const imageHeight = 64;
+        const imageHeight = Math.round(maxHeight / 2);
```
asood123: I'm not following the logic. Why are we dividing by 2, instead of setting the right number up above?

xdamman: Because we are wrapping the image within a SVG of 64px height. And above we get the original image with a height of 128px (to make it retina).

PR #322, opened 2016-06-30
Banner.svg v2

xdamman:
- Added the button become a backer/sponsor
- Making sure we don't display avatars that don't load
- Using caching to avoid having to generate the png every time

### DIFF
--- a/server/controllers/banner.js
+++ b/server/controllers/banner.js
@@ -6,6 +6,8 @@ import _ from 'lodash';
 import Promise from 'bluebird';
 import utils from '../lib/utils';
 
+const requestPromise = Promise.promisify(request, {multiArgs: true});
+
 const filterUsersByTier = (users, tiername) => {
   return _.uniq(filterCollection(users, { tier: tiername }), 'id');
 }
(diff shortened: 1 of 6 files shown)


xdamman:
@asood123 tests keep failing with github:  ✖ Testing if element <body> contains text: "Sign in to GitHub".  - expected "Sign in to GitHub" but got: Skip to content

Can you take a look?

server/controllers/banner.js
```diff
@@ -6,6 +6,8 @@ import _ from 'lodash';
 import Promise from 'bluebird';
 import utils from '../lib/utils';
 
+const requestPromise = Promise.promisify(request, {multiArgs: true});
```
sedubois: can normally just add package `request-promise` and use that directly.

server/lib/utils.js
```diff
@@ -15,9 +15,22 @@ module.exports = {
...
-        fs.writeFileSync(svgFilePath, svg);
-        
-        return svg_to_png.convert(svgFilePath, outputDir)
-                .then(() => readFile(outputFile));
+
+        try {
+            // If file exists, return it
+            // Note: because we generate a md5 fingerprint based on the content of the svg,
+            //       any change in the svg (margin, size, number of backers, etc.) will force
+            //       the creation of a new png :-)
```
sedubois: Is there a mechanism to get rid of old files accumulating on the server? (don't have full picture of how you implemented this)

PR #325, opened 2016-07-04
Updated OnBoardingHeader component so nav links are dynamic

carlosascari:
The logo is customizable

This is a prep for the HomePage design that uses a very similar header, this component can now be reused

Preview of HomePage header:
https://github.com/OpenCollective/OpenCollective/issues/119#issuecomment-227837664


carlosascari:
In light of https://github.com/OpenCollective/OpenCollective/issues/154. 

I will instead create a new header component.

PR #334, opened 2016-07-13
Proxy Collective Header Image

mxstbr:
Can't test because I don't have any data locally, not 100% on the Cloudinary URL config I'm using.


asood123:
copied the PR at https://github.com/OpenCollective/opencollective-website/pull/338 to get the tests to run properly.

PR #337, opened 2016-07-14
first iteration for the opencollective footer banner

xdamman:


asood123:
we should add an e2e test for it that checks that it's serving properly.

PR #344, opened 2016-07-18
Added mailchimp input for waiting list in homepage

carlosascari:
Set overall opacity of homepage bg to 70%
Minor 'css collision' fixes
Added new component homepage/`MailChimpInputSection`

### DIFF
--- a/frontend/src/assets/images/homepage-bg.svg
+++ b/frontend/src/assets/images/homepage-bg.svg
@@ -4,7 +4,7 @@
     <title>FFD4C54E-4AE3-4429-A67B-033D2228B938</title>
     <desc>Created with sketchtool.</desc>
     <defs></defs>
-    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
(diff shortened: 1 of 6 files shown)


xdamman:
Nicely done. It's beautiful.
I've just made a small change to keep it open all the time to increase conversion rate.
Also, FYI I had to change the `componentWillMount` to `componentWillReceiveProps` otherwise `document` would not be set when the page is pre-rendered by the backend.

Great job!

carlosascari:
@xdamman If we include another [MailChimp List Id](https://github.com/OpenCollective/opencollective-website/blob/61348b822df150bd88b88fefb4b94c93f344558a/frontend/src/components/homepage/MailChimpInputSection.js#L9), we can use it to add a second email input under the `Become a sponsor` button, since right now it is also a `mailto:` link. What do you think?

xdamman:
That's a great idea. Here is a list id for sponsors: `4cbda7da19`
Maybe make this one "onclick" so that there is only one visible form by default

xdamman:
I'll merge this once the tests pass so do another PR.

xdamman:
@carlosascari there is one bug: if you hit enter instead of clicking on the button it doesn't work (it returns an error which is annoying)

PR #345, opened 2016-07-19
Smaller CollectiveCard component

carlosascari:
use className `CollectiveCard--large` to revert to original (larger) size, default will be the smaller card

Preview

### DIFF
--- a/frontend/src/css/components/CollectiveCard.css
+++ b/frontend/src/css/components/CollectiveCard.css
@@ -120,7 +120,6 @@
   margin: 15px;
   .CollectiveCard-head {
     width: 100%;
-    height: 220px;
     height: 180px;
   }
   .CollectiveCard-body {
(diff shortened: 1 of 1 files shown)


asood123:
@carlosascari could we reduce the size of the description section as well? The symmetry (where it's equal to the image) or even a little bit smaller looks better. What do you think @erickrico?

carlosascari:
@asood123 
I am hesitant to reduce the font size of the description further, it is now at 14px (from 16px). Anything less, I believe will hurt readability

asood123:
@carlosascari not the font size but the overall space given to description. How many characters can it fit right now?

carlosascari:
Currently the max number of character that fit is about 220
Here is the card reduced in height by ~20px so as to shrink the space in the description
preview

How does that look?

asood123:
LGTM.

PR #346, opened 2016-07-19
Fix links in banner svg

tleunen:
Links were opening inside the svg object, so adding a target to open them in a new tab fixes the issue.

I also fixed the url, to always have a link for every user (either the user's website or the OC slug url), because I believe it was another issue... Some users didn't have urls (@asood123 for example). Let me know if it was expected and I'll revert this part if needed.

### DIFF
--- a/server/controllers/banner.js
+++ b/server/controllers/banner.js
@@ -192,7 +192,7 @@ module.exports = {
         const rawData = responses[i][1];
 
         const contentType = headers['content-type'];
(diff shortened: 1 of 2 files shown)


xdamman:
Brilliant, thank you!

tleunen:
Hmm, there's still a small issue I think. I thought `slug` in `${config.host.website}/${slug}` was the user, but it's the project.

PR #349, opened 2016-07-19
three new eslint rules

asood123:
let over var
brace style is 1tbs
spacing around keywords


asood123:
@carlosascari @sedubois fyi, in case you have comments on this PR or the rules.

Rules are here: https://github.com/OpenCollective/eslint-config-opencollective/blob/master/index.js

xdamman:
1 table spoon of brace style. 
I can get behind that 👌

I like `let` over `var` but in the past, I've been asked to use `var` for consistency with the backend. So just making sure we apply the same rule to api as well.

asood123:
Yep, we can apply that on the backend as well.

asood123:
Ok looks like we can't apply `let` rule to the backend, because that requires `strict-mode` and enabling `strict-mode` breaks a bunch of stuff. I'd still vote for keeping the rule active here and we'll override it on the backend, until we fix stuff that strict-mode breaks. Let me know if anyone disagrees.

PR #350, opened 2016-07-19
Fix sponsor CollectiveCard

carlosascari:
Assert height at 400px
Hide name
Align description vertically 
Hide first footer metric and center align other 2 footer metrics

**Assumptions**
+ A sponsor card does not display their name i.e `CollectiveCard-name`, as the logo should convey this info.
+ In a sponsor card, there are only two entries in the footer i.e `CollectiveCard-footer`, as oppose to the default 3. The **FIRST**, entry is the one that is hidden.

So far, here are the available styles a `CollectiveCard` can become:

preview

### DIFF
--- a/frontend/src/css/components/CollectiveCard.css
+++ b/frontend/src/css/components/CollectiveCard.css
@@ -92,7 +92,7 @@
     }
   }
   &.sponsor {
-    height: 310px;
+    height: 400px !important;
     .CollectiveCard-head {
       height: 160px;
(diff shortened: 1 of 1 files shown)


xdamman:
It looks great on Chrome, but the CSS is a bit broken on Safari
I'd also reduce the height of the description in the classic card and the sponsor card (even more so for sponsors).

Also, can you make the "Start a collective" button from the top bar link to the waiting list (anchor)?

PR #353, opened 2016-07-20
Prevent form from submitting when pressing Enter on input

carlosascari:


xdamman:
Cannot you instead making sure that onSubmit has the same effect than clicking on the button?

carlosascari:
@xdamman onSubmit now has the same effect as clicking on the butotn

xdamman:
Awesome. Thanks!

PR #355, opened 2016-07-20
fix for detecting profile page vs collective page

xdamman:


asood123:
this seems error prone, would be nice to add an e2e test that can show both a group and a profile

PR #357, opened 2016-07-21
add a RelatedGroups component to show on bottom of public page

asood123:


frontend/src/components/CollectiveCard.js
```diff
@@ -29,11 +48,11 @@ export default class CollectiveCard extends Component {
             </div>
             <div className='CollectiveCard-body'>
               <div className='CollectiveCard-name'>{name}</div>
-              <div className='CollectiveCard-description'>{description}</div>
+              <div className='CollectiveCard-description'>We are on a mission to {description}</div>
```
xdamman: Now that this is not only for the homepage, we should move that to the `strings.json` and use `i18n`.

But it's tricky because 
- you should actually use the language of that group and not of that page when you render it.
- you should only show related collectives in the same language.

frontend/src/containers/PublicGroup.js
```diff
@@ -79,7 +80,11 @@ export class PublicGroup extends Component {
             group={group}
             newUserId={newUser.id}
             closeDonationModal={this._closeDonationFlow.bind(this)} />
+          <section className='pt4'>
+            <RelatedGroups title={'Check out similar collectives'} groupList={group.related} />
```
xdamman: we need this in i18n, e.g. after you donate to PartidoDigital or LaPrimaire.org, you want to show text in spanish/french.

frontend/src/components/CollectiveCard.js
asood123: same language is too restrictive right now. Let's come back to it when we have a few hundred active collectives.

asood123:
Found a bug. Don't merge yet.

asood123:
Added a bunch of changes
- serverside rendering for homepage
- fixed an issue that was losing us most of the benefits of ssr
- added logic to not fetch data again after ssr (added a new reducer)
- few other login tweaks

frontend/src/containers/HomePage.js
```diff
@@ -79,7 +52,7 @@ export class HomePage extends Component {
           <div className='heading'>Collect money for your community, transparently.</div>
         </section>
         <section className='HomePageInfo' id="howitworks">
-          <div className='heading'>What is a Open Collective?</div>
+          <div className='heading'>What is an Open Collective?</div>
```
xdamman: I'd argue we should write "an open collective" (or "Open Collective" should refer to the company)

server/routes.js
```diff
@@ -55,15 +55,15 @@ module.exports = (app) => {
   /**
    * Routes
    */
-  app.get('/:slug/banner.md', mw.cache(300), mw.fetchGroupBySlug, mw.fetchUsers(), controllers.banner.markdown);
-  app.get('/:slug/banner.js', mw.cache(3000), mw.fetchGroupBySlug, mw.fetchUsers(), controllers.banner.js);
-  app.get('/:slug/:tier.md', mw.cache(300), mw.fetchGroupBySlug, mw.fetchUsers(), controllers.banner.markdown);
+  app.get('/:slug/banner.md', mw.cache(300), controllers.profile, mw.fetchUsers(), controllers.banner.markdown);
```
xdamman: this should be a middleware, you can only have one controller per route (by design)

server/routes.js
```diff
@@ -72,7 +72,7 @@ module.exports = (app) => {
    * When we refactor PublicGroup to fetch the group in the container, we can remove
    * the explicit routes and just do `app.use(render)`
    */
-  app.get('/', mw.ga, mw.addTitle('OpenCollective - Collect and disburse money transparently'), render);
+  app.get('/', mw.ga, controllers.homepage, mw.addTitle('OpenCollective - Collect and disburse money transparently'), render);
```
xdamman: Middlewares, as their name suggests, should be in the middle. You should end with the controller.

server/routes.js
```diff
@@ -83,12 +83,12 @@ module.exports = (app) => {
   app.get('/github/apply', mw.ga, mw.addTitle('Sign up your Github repository'), render);
   app.get('/connect/github', mw.ga, render);
   app.get('/:slug/connect/twitter', mw.ga, render);
-  app.get('/:slug/edit-twitter', mw.ga, mw.fetchGroupBySlug, render);
+  app.get('/:slug/edit-twitter', mw.ga, controllers.profile, render);
```
xdamman: Ideally, the controller should call render.
(but it was already like that, it's ok to keep it that way for now if you want)

server/views/pages/app.hbs
```diff
@@ -22,7 +22,7 @@
   <body>
     {{> sprite}}
 
-    <div id="content">{{{html}}}</div>
+    <div id="content"><div>{{{html}}}</div></div>
```
xdamman: why this extra div?

asood123: a react artifact that was preventing efficient serverside rendering.

PR #359, opened 2016-07-21
making sure we link to the collectives listed on the profile page

xdamman:


asood123:
Don't merge this. I refactored that code and will make the change as part of the refactor.

PR #360, opened 2016-07-21
make sure we don't load group data on profile page

xdamman:


asood123:
already fixed this as part of refactor. PR coming shortly.

PR #363, opened 2016-07-22
Expenses Redesign

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/165

preview

Donations has also been affected by the redesign

preview

Related:
https://github.com/OpenCollective/OpenCollective/issues/174


asood123:
How does it scale on phone screens?

frontend/src/components/ExpenseForm.js
```diff
@@ -137,32 +112,20 @@ class ExpenseForm extends Component {
...
-              value={attributes.notes}
-              handleChange={notes => appendExpenseForm({notes})} />
-          </div>
-
-          <div className="buttonsRow">
-            <SubmitButton label={i18n.getString('submit')} />
-            <Button color="red" label={i18n.getString('cancel')} onClick={this.onCancel.bind(this)} />
+            
+          <div className='col col-12 sm-col-12 md-col-6 lg-col-6  pl1'>
+            <label>PayPal account/ Account number</label>
```
asood123: I'd s/Paypal Account/Paypal Email. What's the "Account number"?

frontend/src/components/ExpenseForm.js
```diff
@@ -42,93 +31,79 @@ class ExpenseForm extends Component {
...
-            <Input
-              customClass='js-transaction-description'
-              hasError={expense.error.title}
-              value={attributes.title}
-              handleChange={title => appendExpenseForm({title})} />
+        <Notification {...this.props} autoclose/>
+        <div className='line1'>expense details</div>
+        <div className='clearfix input-container'>
+          <div className='col col-12 sm-col-12 md-col-6 lg-col-6 pr1'>
+            <ImageUpload
```
asood123: can you make this a required field? there's an issue already open on that

asood123:
Tested it on mobile - looks great. Can you add a link "Submit another expense" after the expense is submitted.

asood123:
When the user is logged, it's not pre-filling any of the fields. It should prefill email (and not let you edit it). Everything else is fine to let you edit.

frontend/src/components/ExpenseForm.js
carlosascari: The submit button is disabled until it an image is uploaded

frontend/src/components/ExpenseForm.js
carlosascari: It will be removed, so it will only say: `PayPal Email`

PR #364, opened 2016-07-23
support showing all collectives in a supercollective automatically

asood123:
We need a compact version of the collective card to make this manageable. @carlosascari @erickrico

### DIFF
--- a/frontend/src/components/public_group/PublicGroupWhoWeAre.js
+++ b/frontend/src/components/public_group/PublicGroupWhoWeAre.js
@@ -3,10 +3,11 @@ import React from 'react';
 import UserCard from '../../components/UserCard';
 import Markdown from '../../components/Markdown';
 import { displayUrl } from '../../components/DisplayUrl';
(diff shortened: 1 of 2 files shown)


asood123:
Also, we'll need to manually remove members for `opensource` supercollective (I can do that with a query) and review tags for each collectives. Only 36 collectives currently show up for `open source` tag - should be more. @piamancini

piamancini:
do you want me to remove them?

frontend/src/components/public_group/PublicGroupWhoWeAre.js
```diff
@@ -40,6 +41,10 @@ export default class PublicGroupWhoWeAre extends React.Component {
               </div>
             </div>
           ) : null}
+
+          {isSupercollective ?
+            <RelatedGroups title={''} groupList={group.superCollectiveData} {...this.props} />
```
xdamman: That's where the naming breaks. `superCollectiveData` is confusing in this context. It should really be `group.subGroups` or `group.linkedGroups`.

PR #365, opened 2016-07-25
[WIP] Webpack implementation

mxstbr:
Switching to webpack for bundling due to HMR.

**TODO**
- [x] Production bundle
- [x] Development bundle
  - [ ] Hot module replacement

---

There's problems with two modules that actually aren't meant to be used on the client: `joi` and `numbro`. We use them, but they `require` Node.js internal modules like `fs` which don't exist on the client.

I fixed it by spoofing these modules with empty files for webpack, but I'm not sure how your browserify process is working since those aren't actually bundleable. This might've broken email validation (if we're doing that) with `joi`, but I'm not yet 100% sure, would appreciate somebody trying that where we're using it.

_See foretagsplatsen/numbro#74 and hapijs/joi#154 for the solutions I did._

### DIFF
--- a/frontend/config/webpack.prod.babel.js
+++ b/frontend/config/webpack.prod.babel.js
@@ -1,0 +1,38 @@
+var path = require('path');
+var webpack = require('webpack');
+
+module.exports = {
+  entry: [
+    path.join(__dirname, '../src/client.js'),    // Start with src/client.js
+  ],
+  output: {                                      // Bundle everything into dist/bundle.js
+    path: path.resolve(__dirname, '../dist/js'),
+    publicPath: '/',
+    filename: 'bundle.js'
+  },
+  plugins: [
+    new webpack.NormalModuleReplacementPlugin(/^(net|dns)$/, path.resolve(__dirname, 'emptyfile.js')),
+    new webpack.DefinePlugin({                   // Tell webpack if we're in production or not
+      'process.env': {
+        NODE_ENV: JSON.stringify(
+          process.env.NODE_ENV
+        ),
+      },
+    }),
+  ],
+  module: {
+    loaders: [{
+      test: /\.jsx?$/,                             // Transform all .js and .jsx files required somewhere with Babel
+      loader: 'babel',
+      exclude: /node_modules/,
+    }, {
+      test: /\.json$/,
+      loader: 'json',
+    }, {
+      test: /numbro\/numbro/,
+      loader: "imports?require=>false",
+    }],
+  },
+  target: 'web',                                 // Handle global variables of the web, e.g. window
+  progress: true,
+};
--- a/frontend/src/containers/HomePage.js
+++ b/frontend/src/containers/HomePage.js
@@ -48,7 +48,7 @@ export class HomePage extends Component {
               <use xlinkHref='#svg-logotype' fill='#303233'/>
             </svg>
           </div>
-          <div className='subtitle'>organizing the Internet generation</div>
+          <div className='subtitle'>organizing the Internet</div>
           <div className='heading'>Collect money for your community, transparently.</div>
         </section>
         <section className='HomePageInfo' id="howitworks">
--- a/frontend/src/containers/HomePage.js
+++ b/frontend/src/containers/HomePage.js
@@ -48,7 +48,7 @@ export class HomePage extends Component {
               <use xlinkHref='#svg-logotype' fill='#303233'/>
             </svg>
           </div>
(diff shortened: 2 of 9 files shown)


frontend/config/webpack.prod.babel.js
```diff
@@ -0,0 +1,38 @@
+var path = require('path');
+var webpack = require('webpack');
+
+module.exports = {
```
sedubois: note that you could have a single webpack config file as well, and just configure the JSON config object differently depending either on `NODE_ENV` or on `npm_lifecycle_event`. Like here https://github.com/belongapp/belong-website/blob/master/webpack.config.js

mxstbr: Yeah I know, this is a first implementation just to get a feel for it. :grin:

I'll probably do it how we did it for `react-boilerplate` with a `webpack.base.babel.js`: https://github.com/mxstbr/react-boilerplate/tree/master/internals/webpack (the `if/else` in a single file gets unmaintainable when config gets more complex)

frontend/src/containers/HomePage.js
```diff
@@ -48,7 +48,7 @@ export class HomePage extends Component {
               <use xlinkHref='#svg-logotype' fill='#303233'/>
             </svg>
           </div>
-          <div className='subtitle'>organizing the Internet generation</div>
+          <div className='subtitle'>organizing the Internet</div>
```
sedubois: intended?

package.json
```diff
@@ -108,7 +109,8 @@
     "svg-to-png": "^3.1.0",
     "uglify-js": "2.6.2",
     "universal-analytics": "0.3.11",
-    "webpack": "^2.1.0-beta.20"
+    "webpack": "^2.1.0-beta.20",
+    "webpack-dev-server": "^1.14.1"
```
sedubois: better to freeze the dependencies

frontend/src/containers/HomePage.js
mxstbr: nope, accidental 😁 thanks!

package.json
mxstbr: Good catch!

mxstbr:
This is still WIP, dev mode doesn't work because of the server rendering. I might be able to work around that, but I think it's worth talking about disabling SSR in dev maybe? /cc @asood123

package.json
```diff
@@ -39,11 +40,13 @@
     "babel": "5.8.38",
     "babel-core": "6.7.7",
     "babel-eslint": "6.1.2",
+    "babel-loader": "^6.2.4",
```
sedubois: freeze dep and same for others in this file

package.json
```diff
@@ -103,7 +108,9 @@
     "serve-favicon": "2.3.0",
     "svg-to-png": "^3.1.0",
     "uglify-js": "2.6.2",
-    "universal-analytics": "0.3.11"
+    "universal-analytics": "0.3.11",
+    "webpack": "2.1.0-beta.20",
+    "webpack-dev-server": "1.14.1"
```
sedubois: move to dev dependencies?

package.json
```diff
@@ -39,11 +40,13 @@
     "babel": "5.8.38",
     "babel-core": "6.7.7",
     "babel-eslint": "6.1.2",
+    "babel-loader": "^6.2.4",
     "babel-plugin-add-module-exports": "0.1.4",
     "babel-plugin-lodash": "2.3.0",
     "babel-plugin-transform-es2015-modules-commonjs": "6.7.7",
     "babel-preset-es2015": "6.6.0",
     "babel-preset-react": "6.5.0",
+    "babel-preset-react-hmre": "^1.1.1",
```
sedubois: move to dev dependencies?

frontend/src/lib/env.js
```diff
@@ -2,12 +2,14 @@
  * Environment variables
  */
 
-if (typeof process.env.NODE_ENV === 'undefined') {
-  process.env.NODE_ENV = 'development';
+let NODE_ENV = process.env.NODE_ENV;
+
+if (NODE_ENV === 'undefined') {
+  NODE_ENV = 'development';
```
asood123: Doesn't this mean that process.env.NODE_ENV won't be set to anything?

mxstbr: You can't set `process.env.NODE_ENV`, it throws an error and the entire application grinds to halt…

asood123:
What about using this for 'Joi'? https://github.com/jeffbski/joi-browser

PR #368, opened 2016-07-25
Scaled by half, compressed with imagemin

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/190

Shrunk from 229.7kb to 67.8Kb by:
1. Scaling the image by 50% (1322w 838h to 661w 419h)
2. Compressed using [imagemin](https://github.com/imagemin/imagemin)

It looks/renders better as well


asood123:
lgtm. let me figure out why the tests are failing.

PR #369, opened 2016-07-25
Top padding reduced to 5px

carlosascari:
Found a miss alignment, live!
error

Fix
fix


asood123:
merging this even though tests are failing: it's only css and tests are going to take a little bit to fix.

PR #370, opened 2016-07-26
Faq move from website-static repo

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/181


frontend/src/containers/Faq.js
```diff
@@ -0,0 +1,146 @@
...
+      </div>
+    )
+  }
+}
+
+export default connect(mapStateToProps, {})(Faq);
+
+function mapStateToProps() {
+  return {
+    i18n: i18n('en')
```
asood123: we don't need i18n here, fyi

PR #378, opened 2016-08-02
token wasn't being passed properly

asood123:
This fixes `jwt malformed` error that users are seeing on canceling a subscription. We weren't passing the token properly.


asood123:
Merging this to fix production issue. @xdamman @sedubois pls review post-merge and let me know of any issues.

PR #380, opened 2016-08-02
update strings for FreeRideToVote

piamancini:


piamancini:
@asood123 thanks, please let me know when live.

PR #383, opened 2016-08-04
Update /donate page's design

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/201
- Includes Responsive CSS
- `group.mission` collapses Mission line if missing
- `group.host.name` collapses website link/button if missing


xdamman:
The mission seems very very big. Can we reduce it a bit?
Also, doesn't work on mobile

xdamman:
(I pushed small fixes so don't forget to pull first)

PR #393, opened 2016-08-09
Discover

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/198
Requires `discover` branch on **api** repository: https://github.com/OpenCollective/opencollective-api/pull/524

By default shows `All` or any tagged Group
By default it is sorted by `Newest` i.e `ORDER BY group."createdAt" DESC`

*For show options*
Available tags are retrieved from api endpoint `/groups/tags`. 

*For sort options*
Hard-coded to be either Newest or Oldest.

More collectives are displayed when the user scrolls to the bottom of the page, just when the footer begins to show.


frontend/src/reducers/discover.js
```diff
@@ -0,0 +1,27 @@
+import merge from 'lodash/object/merge';
+import { DISCOVER_SUCCESS } from '../constants/discover';
+
+export default function discover(state={}, action={}) {
+  switch (action.type) {
+    case DISCOVER_SUCCESS:
+    const discover = action.json;
+    if (state.show) {
```
asood123: This is pretty hairy. What's the need for unshift?

frontend/src/reducers/discover.js
```diff
@@ -0,0 +1,27 @@
...
+      if (state.show === discover.show && state.sort === discover.sort) {
+        if (state.collectives) {
+          for (let i = state.collectives.length - 1; i >= 0; i--) {
+            discover.collectives.unshift(state.collectives[i]);
+          }
+          return merge({}, state, discover);
+        } else {
+          return merge(state, discover);
```
asood123: I think you always want to merge into a new object, no? Like in line 14.

frontend/src/actions/groups/tags.js
```diff
@@ -0,0 +1,32 @@
...
+    json
+  };
+}
+
+function failure(error) {
+  return {
+    type: constants.GROUP_TAGS_FAILURE,
+    error
+  };
+}
```
asood123: What's the need for this separate action vs. the discover action?

frontend/src/reducers/discover.js
carlosascari: I will explain my thought

`if (state.show === discover.show && state.sort === discover.sort) ...`
When this condition is `true` results are added to the `discover.collectives` array.
This is the only case when paginating via scrolling down.
Otherwise,`discover.collectives` array is replaced with the result (not appened).

If the first condition is `true`, the there is another condition inside:
`if (state.collectives) { ...`
This test is simply to separate the case where `state.collectives` is undefined, which is only true once, every time the page loads.
If `state.collectives` exists, then we are certain that the user is pagination/scrolling.

I will explain the `unshift`.
```
          for (let i = state.collectives.length - 1; i >= 0; i--) {
            discover.collectives.unshift(state.collectives[i]);
          }
```
The current **state** results found in `state.collectives` are prepended to the new results in `discover.collectives`.
This is done so the older or first results come first in the array `discover.collectives` which becomes the new **state** results.

I suppose it simply came naturally to me to do write it this way, I realize now how it can be a bit cryptic. The problem I solved by taking this approach was that when change filters: either  `Show` or `Sort`, old result requested with different filter settings would not be removed from the `state.discover.collectives` array (which is what is rendered), instead the x number of new results would **overwrite** the older results, up-to the size of new results.

**I will explain:**  We have 5 old results, we change the filters and get 3 new results. Out of the 5 old result, the first 3 will be replaced with the 3 new results, the last 2 old results will not be removed or altered.

This happened when using `return merge({}, state, discover);`

frontend/src/reducers/discover.js
carlosascari: Is there a reason why I should always merge into a new object that is, return a new object every time?

As explained above, this is the condition where the page has just loaded and `state.collectives` is undefined.
There I merged the new results directly into the state object that is returned, thats all. 

I don't see why we always create a new object, other than, thats just how its always been?

frontend/src/actions/groups/tags.js
carlosascari: We only need to know the Group tags available to filter by **once** on every page load. 

The Discover action will be called at least once on every page load, however as the user scrolls down new results will be fetched, the Discover action will be called **multiple** times as the user discovers more results with the same filter values. Changing filter values will also trigger the Discover action again.

PR #398, opened 2016-08-16
Added showDisclaimer property for DonationDistributor component

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/199

When set to `true` it will render the disclaimer under the donate button

### DIFF
--- a/frontend/src/components/public_group/DonationDistributor.js
+++ b/frontend/src/components/public_group/DonationDistributor.js
@@ -290,7 +290,7 @@ export default class DonationDistributor extends Component {
 
   render() {
     if (this.props.skipModal) {
-      if (true || this.state.disabled) {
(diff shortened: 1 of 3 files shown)


frontend/src/components/public_group/DonationDistributor.js
```diff
@@ -289,12 +290,13 @@ export default class DonationDistributor extends Component {
 
   render() {
     if (this.props.skipModal) {
-      if (this.state.disabled) {
+      if (true || this.state.disabled) {
```
asood123: why do we need `true ||`? Won't that always just check this.state.disabled?

carlosascari: Good catch, its not suppose to be there, I'll remove it

PR #400, opened 2016-08-16
propose new tagline

piamancini:
I'd like to include the idea of a shared mission and make it about people not about an association.

### DIFF
--- a/frontend/src/containers/HomePage.js
+++ b/frontend/src/containers/HomePage.js
@@ -53,7 +53,7 @@ export class HomePage extends Component {
         </section>
         <section className='HomePageInfo' id="howitworks">
           <div className='heading'>What is an open collective?</div>
-          <div className='subheading'>An unincorporated association that operates in full transparency</div>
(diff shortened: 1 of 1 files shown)


sedubois:
BTW I like Musk's view of business (or lack thereof) - "I don't know what a business is. All a company is is a bunch of people together to create a product or service. There's no such thing as a business, just pursuit of a goal — a group of people pursuing a goal."

http://www.techinsider.io/elon-musk-says-theres-no-such-thing-as-a-business-2015-9

piamancini:
@sedubois I was also playing with the idea of 'shared purpose' instead of mission but since we use mission on our public page, it seemed more fitting

asood123:
👍  for new definition of open collective

sedubois:
I prefer the purpose wording as it's more applicable on an individual level ("life purpose", "inner purpose", etc) and so the collective is just a reflection of the individuals. Mission keeps things a little depersonalized, but it's getting technical, mission works too at this point IMHO

PR #404, opened 2016-08-17
Moved Pending view out from PublicGroup into its own component

carlosascari:
Created `PublicGroupPending` component.


asood123:
Thanks for doing this. Was on my list for next week, once I finish pending-collective backend.

PR #409, opened 2016-08-22
Added public_group components

carlosascari:
Created the following stateless components: 

- `PublicGroupContributors`
- `PublicGroupDonationFlow`
- `PublicGroupExpensesAndActivity`

from refactoring code from the `PublicGroup` container


frontend/src/components/public_group/PublicGroupExpensesAndActivity.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+
+import PublicGroupDonations from './PublicGroupDonations';
+import PublicGroupExpenses from './PublicGroupExpenses';
+
+export default class PublicGroupExpensesAndActivity extends Component {
```
sedubois: can convert to pure functional component

carlosascari: I thought the same at first, however a `{...this.props}` is required and I cannot use it from within a pure functional component

sedubois: what is the problem with the below?

```
export default PublicGroupExpensesAndActivity = props => { 
  ...code...
  <PublicGroupExpenses {...props} />
  <PublicGroupDonations {...props} />
  ...code...
};
```

frontend/src/components/public_group/PublicGroupExpensesAndActivity.js
```diff
@@ -0,0 +1,20 @@
...
+  render() {
+    const { group, expenses, donations, users, itemsToShow } = this.props;
+    return (
+      <section id='expenses-and-activity' className='px2'>
+        <div className='container'>
+          <div className='PublicGroup-transactions clearfix md-flex'>
+            <PublicGroupExpenses group={ group } expenses={ expenses } users={ users } itemsToShow={ itemsToShow } {...this.props} />
```
sedubois: `group`, `expenses` etc are already in `this.props` so specifying them explicitly is redundant normally. `<PublicGroupExpenses {...this.props} />` doesn't work?

frontend/src/components/public_group/PublicGroupExpensesAndActivity.js
carlosascari: Nothing, for some reason I didn't realize I could do that, I will use a pure functional component

frontend/src/components/public_group/PublicGroupExpensesAndActivity.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import PublicGroupDonations from './PublicGroupDonations';
+import PublicGroupExpenses from './PublicGroupExpenses';
+
+export default props => (
```
sedubois: You can also name the component so that its name appears in stack traces/error messages (same comment applies to all functional components):

```
const PublicGroupExpensesAndActivity = props => (...);
export default PublicGroupExpensesAndActivity;
```

PR #416, opened 2016-08-25
[in progress] new route /:slug/expenses/:id/approve

xdamman:
Related to https://github.com/OpenCollective/opencollective-api/pull/551


xdamman:
@carlosascari when you have a chance next week, can you take a look at this?
We can do something nicer than using the default `notify` that shows a top bar (design wise)
Basically, we should convey 3 states: 
- Approving expense (processing)
- Expense approved
- Error while approving expense (with message, typically forbidden)

asood123:
Old. Closing.

PR #417, opened 2016-08-26
add title and separation between core contributors and collectives

asood123:


asood123:
@carlosascari reviewed it. Merging.

PR #418, opened 2016-09-04
form to create a new group and add contributors

asood123:
Note - this has no css. @carlosascari can you review? I'll merge to get it live and you can take a css pass in a separate PR.


frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,396 @@
...
+      mission,
+      description,
+      longDescription,
+      whyJoin,
+      tags,
+      users1,
+      users2,
+      users3,
+      users4,
+      users5
```
asood123: This is hacky. Would be nice to figure out how to support adding up to 20 users without naming them separately. Tried using an array. Ran into issues with merging, etc.

Ideally, there should be two entries and then a "+" button to add more contributors.

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+import createGroup from '../actions/groups/create';
+import notify from '../actions/notification/notify';
+import uploadImage from '../actions/images/upload';
+import validateSchema from '../actions/form/validate_schema';
+
+import groupSchema from '../joi_schemas/group';
+
+import i18n from '../lib/i18n';
+
+const memberRole = 'MEMBER';
```
carlosascari: I've grown accustomed to having `ALL_CAPS` for constants, however I realize that was a pattern intended for es5, what do think?

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+
+export class Addgroup extends Component {
+
+  constructor(props) {
```
carlosascari: I believe the constructor here can be omitted.

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+                  <div className='Addgroup-imagepicker-cont pb3'>
+                    <ImagePicker
+                      uploadOptionFirst
+                      className="logo"
+                      dontLookupSocialMediaAvatars
+                      handleChange={logo => appendGroupForm({logo})}
+                      label="Select collective logo"
+                      uploadImage={uploadImage}
+                      i18n={i18n}
+                      presets={['/static/images/repo.svg', '/static/images/code.svg', '/static/images/rocket.svg']}
```
carlosascari: I'd move the preset arrays to its own constant variable and hoist it above the class for easier future editing, also the array is used again below, so we reuse the Array instead of creating two, with the same elements.

 ```
const LOGO_PRESET = ['/static/images/repo.svg', '/static/images/code.svg', '/static/images/rocket.svg']
```
...
`presets={ LOGO_PRESETS }`
...

File: frontend/src/containers/Addgroup.js [hunk 5]
```diff
@@ -0,0 +1,392 @@
...
+                    />
+                    <ImagePicker
+                      uploadOptionFirst
+                      className="image"
+                      dontLookupSocialMediaAvatars
+                      handleChange={image => appendGroupForm({image})}
+                      label="Select collective image"
+                      uploadImage={uploadImage}
+                      i18n={i18n}
+                      presets={['/static/images/repo.svg', '/static/images/code.svg', '/static/images/rocket.svg']}
```
carlosascari: `presets={ LOGO_PRESETS }`

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+
+    return (
+      <div className='Login'>
+        <Notification {...this.props} />
+        <LoginTopBar />
+        <div className='Addgroup'>
+          <div className='Addgroup-header'>
+            <strong>Create a new collective</strong>
+          </div>
+            <div className="Addgroup-form-container">
```
carlosascari: Double Quote `"` or Single Quote `'`   

I Believe single is the best choice, since it doest require a SHIFT key, and visually it looks slimmer/less.

frontend/src/containers/Addgroup.js
carlosascari: How about an even more hackier hack?

Store using `JSON.stringify()` for the users and a comma separated list for the array:

To Deflate
`[{...},{...},{...}].map( x => JSON.stringify(x) ).toString()`

To Inflate
`"{...},{...},{...}".split().map( x=> JSON.parse(x) )`

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+                        maxLength={255}
+                        placeholder="http://railsgirlsatlanta.com"/>
+
+                      <div className="Addgroup-label">Help us on our mission to...</div>
+                      <CustomTextArea
+                        name={'mission'}
+                        value={mission}
+                        onChange={(value) => appendGroupForm({mission: value})}
+                        maxLength={100}
+                        placeholder="State the core mission of your collective"/>
```
carlosascari: This placeholder is *asking for input*, while most of the others are *showing examples of input*.

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+
+                      <div className="Addgroup-label">Describe your project briefly (short description) </div>
+                      <CustomTextArea
+                        name={'description'}
```
carlosascari: `name='description'`

There is no need to have braces around string literals, there are several of this examples throughout

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+                          value={users5.twitterHandle}
+                          onChange={(value) => appendGroupForm({users5: {name: users5.name, email: users5.email, twitterHandle: value}})}
+                          maxLength={50}
+                          placeholder="@charliechaps"/>
+                      </div>
+                    </div>
+                  </div>
+                </div>
+              </div>
+              <div className={`Addgroup-Button`} onClick={this.create.bind(this)}>create!</div>
```
carlosascari: `.bind(this)` In-lined, inside the `render` function is an anti-pattern, because a **new** function is created on every render.

In the contructor you can bind the method for use:

`this.createRef = this.create.bind(this);`

so now instead you can do:

`... onClick={ this.createRef }`





It is explained here:
https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f#.nsv25mbyg

**tl;dr**
> The anti-pattern is creating new arrays, objects, functions or any other new identities during render or in Redux connect(mapState).

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+      name: attr.name,
+      slug: attr.slug,
+      mission: attr.mission,
+      longDescription: attr.description,
+      logo: attr.logo || '',
+      website: attr.website,
+      data: {
+        utmSource
+      },
+      tags: attr.tags && attr.tags.split(',').map(Function.prototype.call, String.prototype.trim),
```
carlosascari: Never saw `map`, `call` and`trim` used like this, it was intriguing only because I did not understand it at first, what the line does is not obvious. I like it, its clever.

However, this will confuse someone who is not well past novice in JS, I recommend keeping it simple:

`attr.tags.split(',').map(x => x.trim())`

frontend/src/containers/Addgroup.js
```diff
@@ -0,0 +1,392 @@
...
+    return validateSchema(groupForm.attributes, groupSchema)
+      .then(() => createGroup(group))
+      .then(() => this.setState({showConfirmation: true}))
+      .then(() => notify('success', `${group.name} has been created. Refresh page to create a new group.`))
+      .catch(({message}) => notify('error', message));
+  }
+
+}
+
+
```
carlosascari: Unnecessary line break (one is fine)

frontend/src/joi_schemas/group.js
```diff
@@ -0,0 +1,25 @@
+import Joi from 'joi';
+
+const schema = Joi.alternatives().try(
+  Joi.object().keys({
+    name: Joi.string().label('Name').required(),
+    slug: Joi.string().label('Slug').required(),
+    currency: Joi.string().label('Currency'),
+    website: Joi.string().uri().allow('').label('Website url'),
```
carlosascari: **NOTICE**

If I remember right, `Joi.uri()` Will correctly (by the RFC standards) invalidate a url without a **protocol**: 

So while `http://google.com` is valid, `google.com` is not. 

Something to keep in mind.

carlosascari:
Review completed, I did not find bugs however there are some minor improvements that can be made.

frontend/src/joi_schemas/group.js
asood123: yep, good reminder. we should think of an alternative.

frontend/src/containers/Addgroup.js
asood123: fixed.

frontend/src/containers/Addgroup.js
asood123: fixed

frontend/src/containers/Addgroup.js
asood123: good catch. fixed.

frontend/src/containers/Addgroup.js
asood123: That's interesting. Something for next pass.

PR #421, opened 2016-09-05
switch image uploads to text fields & support avatar

asood123:


frontend/src/containers/AddGroup.js
```diff
@@ -198,27 +186,37 @@ export class AddGroup extends Component {
                          <Input
                           name='email'
                           value={users1.email}
-                          onChange={(value) => appendGroupForm({users1: {name: users1.name, email: value, twitterHandle: users1.twitterHandle}})}
+                          onChange={(value) => appendGroupForm({users1: {name: users1.name, email: value, twitterHandle: users1.twitterHandle, avatar: users1.avatar}})}
                           maxLength={50}
```
carlosascari: I've noticed several fields are limited to 50 characters, this *may* one day be an issue. An **email** for example can have up to [254 characters](https://en.wikipedia.org/wiki/Email_address#Syntax)

Are you sure you want to limit it to 50?

carlosascari:
@asood123  LGTM

asood123:
merging this to unblock us.

frontend/src/containers/AddGroup.js
asood123: I suspect that we'll change this flow long before then ;-)

xdamman:
What's the rationale for doing this?

PR #423, opened 2016-09-06
(on going) Editable collective

carlosascari:
Currently a user is able to edit:
 - Name
 - Logo
 - Background Image
 - Mission
 - Description
 - Long Description
 - Website

To view the editor, simply go the route `opencollective.com/:slug/edit`. You must be logged in, in order to save changes.

You can make a change and preview it before saving, as changes will only be saved when clicking the **Publish** button.


asood123:
Where is the `edit` mode triggered from? And what auth is in place to only allow it for people who are `MEMBERS` of that collective?

carlosascari:
@asood123 
- `edit` mode is triggered by going to the url: `opencollective.com/:slug/edit`, there is no link in place as of now, however there will be one on the `LoginTopBar` menu only when a user is logged in.

- Currently there is no auth to only allow it for people who are members of the collective. 
Yesterday I worked with the `api` repo having the *group update endpoint* include a auth middleware: 
`aZ.authorizeAccessToGroup({userRoles: [HOST, MEMBER]})`
https://github.com/OpenCollective/opencollective-api/blob/86a8a719e2df9f5da3d10d5e170e539b7d73cbd4/server/routes.js#L142`
However, I can only see that after a merge, it is now missing:
https://github.com/OpenCollective/opencollective-api/blob/1eb5362853f6a276e93af2a6ba339880a982a250/server/routes.js#L143

I am assuming that removing the auth was unintentional? If that is the case then we can add the auth middleware back to the endpoint.

asood123:
@carlosascari i'll take care of the auth

frontend/src/containers/EditCollective.js
```diff
@@ -0,0 +1,370 @@
...
+    fields[highlightField] = newValue;
+    this.setState({fields: fields});
+    this.updateHighlights();
+  }
+
+  onClickHighlight(highlight) {
+    this.setState({
+      showModal: true,
+      highlightLabel: highlight.label,
+      highlightField: typeof highlight.field === 'funciton' ? highlight.field(this.state) : highlight.field,
```
sedubois: typo `funciton` will fix in https://github.com/OpenCollective/opencollective-website/pull/424 (where the error is spotted by new eslint rules)

frontend/src/containers/EditCollective.js
```diff
@@ -0,0 +1,370 @@
...
+      </svg>
+    </div>
+    <div className='EditCollective-TopBar-buttons'>
+      <div className='EditCollective-TopBar-Button' onClick={ props.onPublish }>Publish</div>
+      <a href='.'><div className='EditCollective-TopBar-Button trans'>Exit Edit Mode</div></a>
+    </div>
+  </div>
+);
+
+export default class EditCollective extends Component {
```
sedubois: In future please try to separate into presentational and container components (separate files). A component is either pure or performs side-effects such as sync with Redux, notify, hold a state, read globals, etc.

https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.kgdpj3cfy
https://medium.com/building-asana/designing-simpler-react-components-13a0061afd16#.mnvo638yp

carlosascari: Will do

PR #424, opened 2016-09-06
Eslint website

sedubois:
Pushing the stuff done so far (e.g doesn't have no-commonjs rule)


frontend/src/reducers/form.js
```diff
@@ -42,29 +41,34 @@ const profileInitialState = {
 
 function profile(state=profileInitialState, action={}) {
   switch (action.type) {
-    case constants.SET_EDIT_MODE_PROFILE:
-      if (!action.isEditMode) {
-        return merge({}, profileInitialState, { isEditMode: action.isEditMode });
-      }
-      return merge({}, state, { isEditMode: action.isEditMode });
+    // TODO constants.SET_EDIT_MODE_PROFILE doesn't exist, should it?
```
asood123: you can remove all of the constants from this reducer that don't exist. `APPEND_PROFILE_FORM` is the only one we are using.

frontend/src/reducers/session.js
```diff
@@ -21,8 +21,9 @@ export default function session(state={
         isAuthenticated: false
       });
 
-    case constants.SHOW_POPOVERMENU:
-      return merge({}, state, { hasPopOverMenuOpen: action.hasPopOverMenuOpen });
+    // TODO constants.SHOW_POPOVERMENU doesn't exist, should it?
```
asood123: nope. pls remove. We copied some of this code from `app` when we originally split the two. leftovers.

package.json
```diff
@@ -12,7 +12,7 @@
     ]
   },
   "scripts": {
-    "start": "NODE_CONFIG_DIR=./server/config node server",
+    "start": "NODE_CONFIG_DIR=./server/config babel-node server",
```
asood123: I thought we aren't supposed to use `babel-node` in production?

package.json
```diff
@@ -58,6 +58,9 @@
     "cssnano": "3.5.2",
     "ejs": "2.4.1",
     "envify": "3.4.0",
+    "eslint-plugin-import": "^1.14.0",
```
asood123: can you freeze all of these?

package.json
sedubois: Will fix this in next PR https://github.com/OpenCollective/opencollective-website/pull/427

PR #425, opened 2016-09-06
Lodash update

datapimp:
This updates the project's version of lodash from a 3.x version to the latest 4.x version.  

The main difference is that the lodash project has moved away from the `lodash/object/module` style of import to the more flattened `lodash/module` style.  I updated all of our imports to be inline with this change.

There are also a couple of deprecations -- namely `pluck`, which is replaced by map;  It has the same exact argument signature / behavior.  

And the `contains` function which we use to test for the existence of a pattern in a string.  I went with using a regex on the value, after ensuring it is a string, which should provide the same results.

Please note the babel plugins were also updated in this PR, as we use the lodash babel plugin I went ahead and updated them all to their latest versions


package.json
```diff
@@ -36,17 +36,17 @@
...
-    "babel-eslint": "6.1.2",
-    "babel-plugin-add-module-exports": "0.1.4",
-    "babel-plugin-lodash": "2.3.0",
-    "babel-plugin-transform-es2015-modules-commonjs": "6.7.7",
-    "babel-preset-es2015": "6.6.0",
-    "babel-preset-react": "6.5.0",
-    "babel-preset-stage-0": "6.5.0",
-    "babel-register": "6.7.2",
-    "babelify": "7.2.0",
+    "babel-cli": "^6.14.0",
```
asood123: can you freeze all of these?

datapimp: no problem; just froze all of the dependencies in the `package.json` file

PR #429, opened 2016-09-07
store video on the backend

asood123:


test/server/banner.spec.js
```diff
@@ -1,11 +1,12 @@
...
+import sinon from 'sinon';
+import sizeOf from 'image-size';
 
 import api from '../../server/src/lib/api';
 import app from '../../server/src/index';
-import request from 'supertest';
-import sizeOf from 'image-size';
 
-import sinon from 'sinon';
+import shieldIONock from '../data/shields.io.nock'; // eslint-disable-line
```
sedubois: why is the disable-line necessary?

test/server/banner.spec.js
```diff
@@ -89,6 +90,8 @@ describe("badge", () => {
     });
   });
 
+  beforeEach(() => shieldIONock());
```
sedubois: can also write `beforeEach(shieldIONock);`

test/server/banner.spec.js
asood123: not sure why i added that. removed.

test/server/banner.spec.js
asood123: changed.

PR #433, opened 2016-09-13
Editable Collective - Members

carlosascari:
Can preview removal, updates and created members before saving changes.

@asood123
There is a new `updateMembers` action. It sends a PUT request to `/groups/${groupid}/members`

It is intended to send an Array with Objects.
Each Object will have: `id`, `name`, `avatar`, `core` properties. 

If `id` is missing, it is understood that it is a new member that is being added to the group.
Otherwise if there is a `id`, the member is being updated.


frontend/src/actions/groups/updateMembers.js
```diff
@@ -0,0 +1,43 @@
+import { putJSON } from '../../lib/api';
+import * as constants from '../../constants/groups';
+
+/**
+ * Update group members
+ */
+
+export default (groupid, group) => {
+  const url = `/groups/${groupid}/members`;
```
asood123: Does this call already exist on the backend?

carlosascari: No id does not, sorry I thought I mentioned it above

PR #446, opened 2016-09-21
Webpack 2 asset paths

asood123:


frontend/src/store/middleware.js
```diff
@@ -0,0 +1,17 @@
...
+const options = {};
+
+if (process.env.NODE_ENV === 'production') {
+  // Only show failure on production
+  options.predicate = (getState, action) => {
+    return `${action.type}`.match(/FAILURE/)
+  };
+}
+
+//const logger = createLogger(options);
```
asood123: Do we need to uncomment this? Or remove `createLogger` import?

PR #451, opened 2016-09-22
removed app link from login top bar

asood123:


asood123:
doing this part of larger change. Closing for now.

PR #461, opened 2016-09-25
splitting firstName and lastName after donation

xdamman:
we only ask for name, but we automatically split it in firstName and lastName

### DIFF
--- a/frontend/src/containers/PublicGroup.js
+++ b/frontend/src/containers/PublicGroup.js
@@ -301,7 +301,13 @@ export function saveNewUser() {
   } = this.props;
 
   return validateSchema(profileForm.attributes, profileSchema)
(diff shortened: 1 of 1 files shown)


frontend/src/containers/PublicGroup.js
```diff
@@ -301,7 +301,13 @@ export function saveNewUser() {
   } = this.props;
 
   return validateSchema(profileForm.attributes, profileSchema)
-    .then(() => updateUser(newUser.id, profileForm.attributes))
+    .then(() => {
+      const userData = Object.assign({}, profileForm.attributes);
+      const nameTokens = userData.name.split(' ');
+      userData.firstName = nameTokens.shift();
+      userData.lastName = nameTokens.join(' ');
+      updateUser(newUser.id, userData)
```
asood123: needs a return statement.

PR #469, opened 2016-09-26
Support for multipage in longDescription

xdamman:


frontend/src/components/public_group/PublicGroupHero.js
```diff
@@ -33,6 +34,12 @@ export default class PublicGroupHero extends Component {
   render() {
     const { group, i18n, session, hasHost, canEditGroup, groupForm, appendEditGroupForm} = this.props;
 
+    const titles = Object.keys(processMarkdown(group.longDescription));
+
+    const getAnchor = (title) => {
+      return title.toLowerCase().replace(' ','-').replace(/[^a-z0-9\-]/gi,'')
```
carlosascari: The `toLowerCase()` call removes the need to make the regex case insentive (you can remove the `i`)

test/unit/lib/process_markdown.js
```diff
@@ -0,0 +1,19 @@
+import {expect} from 'chai';
+import processMarkdown from '../../../frontend/src/lib/process_markdown';
+
+import mdfile from '../../data/brusselstogether.md.json';
+
+describe('processMarkdown', () => {
+
+  it('should return the initial text if no title', () => {
+    const text = "Hello world";
```
carlosascari: This string should be in single quotes for consistency

PR #470, opened 2016-09-28
Edit profile

xdamman:


frontend/src/components/ContentEditable.js
```diff
@@ -10,7 +11,10 @@ export default class ContentEditable extends React.Component {
   }
 
   render() {
-    const { tagName, html, ...props } = this.props;
+    const { format, tagName, html, ...props } = this.props;
+
+    if (format === 'markdown' && this.props.disabled)
```
carlosascari: You can add `disabled` in the list of props: `const { format, tagName, html, ...props } = this.props;` line 14
So as to keep the props that are used in the function, grouped together

BTW @xdamman what magic is this: `{ ...props } = this.props`
Where is `props` defined I only see `this.props` available

PR #474, opened 2016-09-30
Contributors

xdamman:
Mixing core contributors and backers into "members" (but showing first the core contributors).

The idea here is that we want to promote the concept of flat organization and fluid leadership. Everyone is a contributor and every contributor is a member of the collective. A backer/sponsor is also a contributor. A contributor on Github is a contributor. And anyone can become a core contributor.

With a lot of contributors (more than 100)


piamancini:
I like this!

package.json
```diff
@@ -97,6 +97,7 @@
     "react-onclickoutside": "4.9.0",
     "react-redux": "4.4.5",
     "react-router": "1.0.3",
+    "react-sticky": "^5.0.5",
```
asood123: can you remove `^`?

PR #477, opened 2016-09-30
Added border and white bg for hover and focus states

carlosascari:
When hover over a editable component, it will be surrounded by a dotted border and will have a white background. 
The background and border will persist if the component has focus.

### DIFF
--- a/frontend/src/css/components/ContentEditable.css
+++ b/frontend/src/css/components/ContentEditable.css
@@ -13,10 +13,12 @@ span.editing {
   border: 1px dotted transparent;
   &.ContentEditable--enabled:hover {
     background: white;
+    color: black;
     border: 1px dotted black;
   }
   &.ContentEditable--enabled:focus {
     background: white;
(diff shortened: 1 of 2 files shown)


frontend/src/containers/ProfilePage.js
```diff
@@ -63,7 +63,8 @@ export class ProfilePage extends Component {
               format='markdown'
               disabled={ !canEditUser }
               onChange={ event => updateUser(profile.id, {description: event.target.value}) }
-              placeholder={i18n.getString('defaultDescription')} />
+              placeholder={i18n.getString('defaultDescription')}
+              style={{textAlign: 'center'}} />
```
xdamman: shouldn't this be in the .css?

xdamman:
It looks nice but it doesn't work for the mission because the text is in white:

xdamman:
@carlosascari FYI I fixed it by setting the color of the text to black

PR #481, opened 2016-10-03
Using group.slug instead of group.id to simplify things

xdamman:
Slug is also a unique idea but it has the benefit of always being available in the URL.

This also makes it easier to handle the case of the redirect after login with `pushState`.
Today, it doesn't work and it shows a group page that is empty (the user needs to hard refresh the page). That's because the current page changes but it didn't get the opportunity to do a call to the API to turn the slug into a group id. We could handle that particular case differently but I figured it would be more elegant to use the `group.slug` everywhere. It's also easier to debug as a slug is more human readable than an id.

It also simplifies the custom categories for all wwcode groups and others.

I also took the opportunity to make our code works with the Redux Google Chrome Extension. It's pretty cool and useful to debug step by step.

### DIFF
--- a/frontend/src/components/public_group/PublicGroupHero.js
+++ b/frontend/src/components/public_group/PublicGroupHero.js
@@ -33,7 +33,6 @@ export default class PublicGroupHero extends Component {
 
   render() {
     const { group, i18n, session, hasHost, canEditGroup, groupForm, appendEditGroupForm} = this.props;
-
     const titles = Object.keys(processMarkdown(group.longDescription));
 
     const getAnchor = (title) => {
@@ -55,16 +54,18 @@ export default class PublicGroupHero extends Component {
         <div className='container relative center'>
           <LoginTopBar loginRedirectTo={ `/${ group.slug }` } />
           <div className='PublicGroupHero-content'>
-            <UserPhoto
-              editable={canEditGroup}
-              onChange={logo => {
-                if (logo !== group.logo)
-                  return appendEditGroupForm({logo})
-              }}
-              user={{ avatar: groupForm.attributes.logo || group.logo }}
-              className='PublicGroupHero-logo mb3 bg-contain'
-              presets={[]}
-              {...this.props} />
+            {group.logo &&
+              <UserPhoto
+                editable={canEditGroup}
+                onChange={logo => {
+                  if (logo !== group.logo)
+                    return appendEditGroupForm({logo})
+                }}
+                user={{ avatar: groupForm.attributes.logo || group.logo }}
+                className='PublicGroupHero-logo mb3 bg-contain'
+                presets={[]}
+                {...this.props} />
+            }
 
             <p ref='PublicGroupHero-name' className='PublicGroup-font-20 mt0 mb2'>{ i18n.getString('hiThisIs') }
               <a href={ group.website }> { group.name }</a> { i18n.getString('openCollective') }.
--- a/frontend/src/containers/PublicGroup.js
+++ b/frontend/src/containers/PublicGroup.js
@@ -3,7 +3,6 @@ import { connect } from 'react-redux';
 
 import sortBy from 'lodash/sortBy';
 import take from 'lodash/take';
-import values from 'lodash/values';
 import merge from 'lodash/merge';
 
 import filterCollection from '../lib/filter_collection';
--- a/frontend/src/containers/PublicGroup.js
+++ b/frontend/src/containers/PublicGroup.js
@@ -17,7 +17,7 @@ import appendDonationForm from '../actions/form/append_donation';
 import appendProfileForm from '../actions/form/append_profile';
 import decodeJWT from '../actions/session/decode_jwt';
 import donate from '../actions/groups/donate';
-import fetchGroup from '../actions/groups/fetch_by_id';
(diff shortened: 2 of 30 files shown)


frontend/src/components/public_group/PublicGroupHero.js
```diff
@@ -55,16 +54,18 @@ export default class PublicGroupHero extends Component {
...
-              editable={canEditGroup}
-              onChange={logo => {
-                if (logo !== group.logo)
-                  return appendEditGroupForm({logo})
-              }}
-              user={{ avatar: groupForm.attributes.logo || group.logo }}
-              className='PublicGroupHero-logo mb3 bg-contain'
-              presets={[]}
-              {...this.props} />
+            {group.logo &&
```
asood123: Won't this check prevent someone from adding a logo when editing a collective if there isn't one?

frontend/src/containers/PublicGroup.js
```diff
@@ -201,28 +200,28 @@ export class PublicGroup extends Component {
...
-    return Promise.all([
-      fetchGroup(group.id),
-      fetchTransactions(group.id, FETCH_DONATIONS_OPTIONS),
-      fetchTransactions(group.id, FETCH_EXPENSES_OPTIONS),
-      fetchUsers(group.id)
-    ])
+    if (!group.name) promises.push(fetchGroup(slug));
+    if (!group.usersByRole) promises.push(fetchUsers(slug));
+    if (!transactions) promises.push(fetchTransactions(slug, FETCH_DONATIONS_OPTIONS));
+    if (!transactions) promises.push(fetchTransactions(slug, FETCH_EXPENSES_OPTIONS));
```
asood123: previous two can be combined in one line.

frontend/src/containers/PublicGroup.js
```diff
@@ -201,28 +200,28 @@ export class PublicGroup extends Component {
...
+
+    Promise.all(promises).then();
```
asood123: don't think .then() here. Promise.all() should be enough.

frontend/src/containers/PublicGroup.js
```diff
@@ -248,25 +243,26 @@ export class PublicGroup extends Component {
...
       fetchTransactions
     } = this.props;
 
     return Promise.all([
-      fetchGroup(group.id),
-      fetchUsers(group.id),
-      fetchTransactions(group.id, FETCH_DONATIONS_OPTIONS)
+      fetchProfile(slug),
+      fetchTransactions(slug, FETCH_DONATIONS_OPTIONS),
+      fetchTransactions(slug, FETCH_EXPENSES_OPTIONS),
```
asood123: why refetch expenses here?

frontend/src/ui/expenseCategories.js
```diff
@@ -1,67 +1,16 @@
...
-export default function(groupid) {
-  if (typeof groupid == "string")
-    groupid = parseInt(groupid, 10);
+export default function(slug = '') {
 
   let categories = [];
-  switch (groupid) {
+  let base_slug = slug;
+  if (slug.match(/^wwcode/))
+    base_slug = 'wwcode';
```
asood123: nice!

frontend/src/containers/PublicGroup.js
xdamman: No it's one for donations and one for expenses, I'm refactoring that in the coming PR.

asood123: oh i meant, it's two things with the same if statement. It could be `if (!transactions) promises.push(fetchTransactions(..), fetchTransactions(..))`

frontend/src/components/public_group/PublicGroupHero.js
xdamman: Mmmm... It might. The problem is that if I don't do that, the child component doesn't update once a value is fetched from the API (after we login). One way we could solve it is by returning a default logo from the API. What do you think?

frontend/src/containers/PublicGroup.js
xdamman: I've removed this in the refactoring of the next PR

frontend/src/containers/PublicGroup.js
xdamman: Good point. I'll take of that in the new PR if you don't mind, because I got rid of `fetchTransactions`

frontend/src/containers/PublicGroup.js
xdamman: gotcha. Well, I got rid of `fetchTransactions` anyway in the coming PR, so let's ignore this for now.

frontend/src/components/public_group/PublicGroupHero.js
asood123: Which child component doesn't update? I saw some issues like that with ImagePicker. 

Many collectives don't have a logo and it's nice to keep layout clean.

frontend/src/containers/PublicGroup.js
asood123: 👍

frontend/src/containers/PublicGroup.js
asood123: 👍

PR #483, opened 2016-10-03
Updated WWCode strings

piamancini:


asood123:
only e2e tests are failing due to backend issues. merging.

PR #486, opened 2016-10-04
Allow slug to have - or _

xdamman:
Fixes https://github.com/OpenCollective/OpenCollective/issues/253


asood123:
tried merging master to this and something went wonky with git. won't let me push to it at all. created a new PR for this change.

PR #491, opened 2016-10-07
allows editing of long description in markdown

asood123:
Added editable state to our markdown component.

It starts out as showing markdown. onClick will turn it editable. onBlur will switch back to markdown


xdamman:
tests don't pass

PR #493, opened 2016-10-07
show create collective button

asood123:


xdamman:
I don't think we are ready yet.

This is how the collective page looks like once the creator click on the "confirm" button in the email: https://opencollective.com/xdamman-test

PR #494, opened 2016-10-07
only show backers on widget

asood123:


xdamman:
sorry I didn't know you were fixing this as well.
So I already did it here: https://github.com/OpenCollective/opencollective-website/pull/495

PR #510, opened 2016-10-15
return a promise

asood123:


asood123:
merging to unblock production. @xdamman pls review post-merge.

PR #511, opened 2016-10-15
remove git clean from deploy scripts

asood123:


asood123:
merging without tests completing to get production fixed.

PR #513, opened 2016-10-18
New collective page

asood123:


asood123:
Tests keep timing out... Can't figure out why. They pass locally and when I ssh into circleci.

frontend/src/containers/Collective.js
```diff
@@ -0,0 +1,273 @@
+import React, { Component, PropTypes } from 'react';
```
xdamman: The problem with adding new files, is that it's impossible for me to see what you changed in them.
E.g. How different is this page compared to the former PublicGroupPage.js?

frontend/src/reducers/collectives.js
```diff
@@ -0,0 +1,56 @@
...
+          [action.data.collective.slug]: merge({}, {settings: DEFAULT_COLLECTIVE_SETTINGS}, action.data.collective)
+        });
+
+      }
+      return state;
+
+    case constants.GROUP_SUCCESS:
+      return merge({}, state, action.groups);
+
+    // 8: {
```
xdamman: Can remove those comments

frontend/src/reducers/expenses.js
```diff
@@ -3,6 +3,7 @@ import * as constants from '../constants/expenses';
 
 export default function expenses(state={}, action={}) {
   switch (action.type) {
+    case constants.PENDING_EXPENSES_SUCCESS:
```
xdamman: it would be more readable to add the verb at the beginning of those constants, e.g. `FETCH_PENDING_EXPENSES_SUCCESS`

frontend/src/selectors/collectives.js
```diff
@@ -0,0 +1,106 @@
...
+        settings,
+        { host },
+        { members },
+        { backers },
+        { backersCount: backers.length },
+        { contributors },
+        { contributorsCount: contributors.length },
+        { expenses: filterCollection(expenses, { GroupId: collective.id })},
+        { transactions: filterCollection(transactions, {GroupId: collective.id })},
+    ));
```
xdamman: I have mixed feeling with those selectors. I understand their utility but it's adding complexity. Is it worth it for us at our stage?

I'm all for giving it a try, but you'll have to show me more next week.

frontend/src/components/UserAvatarRow.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default class UserAvatarRow extends Component {
+
+  render() {
+    const {
+      backers,
+      members
+    } = this.props;
+
+    const all = unionBy(members, backers, 'username');
```
asood123: yes

frontend/src/containers/Collective.js
asood123: Honestly, that doesn't really matter in this PR. This is setting up structure and adding some components that carried over. Once it's merged, we can modify individual components.

frontend/src/selectors/collectives.js
asood123: It's sort of a wash for us. Pros: they are way cleaner and calls like `filterCollection` shouldn't be in `mapStateToProps` because they get called all the time. That being said, our dataset is small, so it may not matter that much. 

Not much to learn with them, so I'd say we try them with this iteration. If it becomes an issue, we can switch back.

PR #517, opened 2016-10-22
add an unknown check to allow other fields

asood123:


asood123:
merging to get production unblocked. @xdamman pls review post-merge.

PR #524, opened 2016-10-28
fix sponsors on collective-card

asood123:


asood123:
merging to fix production issue. @xdamman

PR #528, opened 2016-10-31
don't show twice a member also backer

xdamman:


xdamman:
Closing this as I have a better fix in https://github.com/OpenCollective/opencollective-website/pull/530

PR #530, opened 2016-10-31
show number of sub collectives, don't show backers if > 50

xdamman:


frontend/src/ui/strings.json
```diff
@@ -342,6 +348,7 @@
    "it": {
     "activitiesPlaceholderText": "Condividi questa pagina per trovare sostenitori e finanziatori che vogliano supportare il tuo collettivo!",
     "activitiesPlaceholderTitle": "Le donazioni verranno visualizzate qui",
+    "across": "",
```
asood123: 👍  I like this as a reminder that we need to translate. Though maybe it's better to populate with the english version? Otherwise, frontend will skip that word and we may not notice.

PR #541, opened 2016-11-03
fix for cents in sponsor card

xdamman:


asood123:
Merging without waiting for tests to fix production

PR #544, opened 2016-11-07
Mosaic for members and backers

xdamman:

### DIFF
--- a/frontend/src/components/collective/CollectiveMembers.js
+++ b/frontend/src/components/collective/CollectiveMembers.js
@@ -1,8 +1,20 @@
 import React, { PropTypes } from 'react';
-import filterCollection from '../../lib/filter_collection';
 import UserCard from '../../components/UserCard';
+import Mosaic from '../../components/Mosaic';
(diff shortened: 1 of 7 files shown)


frontend/src/components/collective/CollectiveMembers.js
```diff
@@ -1,28 +1,39 @@
...
 
+  constructor(props) {
+    super(props);
+    const { group } = props;
+    this.contributors = _.uniqBy(_.union(group.members, group.backers.filter(b => !b.tier.match(/sponsor/i))), 'id');
+    this.contributors = this.contributors.map(c => {
+      c.href = `/${c.username}`;
+      return c;
+    });
+    this.sponsors = _.uniqBy(group.backers.filter(b => b.tier.match(/sponsor/i)), 'id');
```
asood123: since this is in the constructor, will it update after a donation?

xdamman: Nope. Does it matter? I can put that in the render but it's less efficient.

PR #548, opened 2016-11-08
Fix CSS for Edit Collective TopBar

xdamman:


asood123:
merging since it's css only.

PR #564, opened 2016-11-23
making sure we don't crash if we can't load sponsor image

xdamman:


xdamman:
merging and deploying to avoid more crashing

PR #579, opened 2016-12-02
disable login after first click

asood123:


frontend/src/containers/Login.js
```diff
@@ -20,7 +20,10 @@ import Confirmation from '../components/Confirmation';
 export class Login extends Component {
   constructor(props) {
     super(props);
-    this.state = { showConfirmation: false };
+    this.state = { 
+      showConfirmation: false,
+      inProgress: false,
```
xdamman: isn't `inProgress` enough? i.e. `disabled={!this.state.inProgress}`

asood123: yep, fixed

PR #580, opened 2016-12-02
reduce double clicking on expense form

asood123:


frontend/src/components/ExpenseForm.js
```diff
@@ -163,7 +164,14 @@ class ExpenseForm extends Component {
           )}
 
         </div>
-        <button type='submit' className={`Button ${hasAttachment ? 'Button--green' : 'Button--disabled'}`} onClick={hasAttachment && this.onSubmit.bind(this)}>submit expense</button>
+        <AsyncButton
+          color='green'
+          inProgress={expense.inProgress}
+          disabled={expense.inProgress}
```
xdamman: the button should be smart enough to know that it should be disabled while in progress.

asood123: hah, actually it is

PR #582, opened 2016-12-03
process SVG if we need to transform it

xdamman:


server/src/lib/utils.js
```diff
@@ -11,7 +11,7 @@ export function getCloudinaryUrl(src, { width, height, query }) {
   const cloudinaryBaseUrl = 'https://res.cloudinary.com/opencollective/image/fetch';
 
   // We don't try to resize animated gif, svg or images already processed by cloudinary
-  if (src.substr(0, cloudinaryBaseUrl.length) === cloudinaryBaseUrl || src.match(/\.gif$/) || (src.match(/\.svg/)) || src.match(/localhost\:3000/)) {
+  if (src.substr(0, cloudinaryBaseUrl.length) === cloudinaryBaseUrl || src.match(/\.gif$/) || (src.match(/\.svg/) && !query) || src.match(/localhost\:3000/)) {
```
xdamman: we usually don't want to send .svg images through cloudinary because there is no point in resizing a SVG...
... unless we want to apply transformations such as adding a circle of color, like I'm now doing in the new monthly email template, see https://github.com/OpenCollective/opencollective-api/pull/698
(we could technically add such circle in SVG but email clients don't always support SVG)

PR #599, opened 2016-12-27
if we don't need to transform the image, no need to proxy it

xdamman:
Fix for https://github.com/OpenCollective/OpenCollective/issues/275


asood123:
weren't we proxying it to reduce costs at cloudinary?

xdamman:
see the code, this is not going through cloudinary.

basically if imageUrl is a `.svg` and we don't need to transform it (like adding a circle around), then there is no need to resize it through cloudinary and proxy it. We can directly serve it from where it is (in this case our own server in `/static/images/rocket.svg` but it could also be on a 3rd party server)

PR #600, opened 2016-12-28
new route to approve/reject expense automatically

xdamman:
New route `/:slug/expenses/:expenseid/:action(approve|reject)` to automatically approve or reject an expense. The API will return an error if the logged in user doesn't have the right permissions.

### DIFF
--- a/frontend/src/containers/Ledger.js
+++ b/frontend/src/containers/Ledger.js
@@ -39,7 +39,7 @@ import { getAppRenderedSelector } from '../selectors/app';
 import { 
   getUsersSelector } from '../selectors/users';
 import { getAuthenticatedUserSelector } from '../selectors/session';
-import { getPathnameSelector } from '../selectors/router';
+import { getPathnameSelector, getParamsSelector } from '../selectors/router';
 import {
   getApproveInProgressSelector,
   getRejectInProgressSelector,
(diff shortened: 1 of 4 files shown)


frontend/src/containers/Ledger.js
```diff
@@ -157,9 +159,19 @@ export class Ledger extends Component {
...
 
+    switch (params.action) {
+      case 'approve':
+        this.approveExp(params.expenseid);
+        break;
+      case 'reject':
+        this.rejectExp(params.expenseid);
+        break;
+    }
+
```
asood123: Where is the confirmation message to let the user know that expense was approved or rejected?

xdamman: It's using the same logic then when the user clicks on the "approve" button.
The user will see the expense with the label "approved" in the list of pending expenses to be paid at the top.

asood123: Ok. Fine for now. Something like `notify('success', 'Expense approved')` in the approve promise chain would be useful.

xdamman: Sure but then it should be added within the `approveExp` function.

asood123: yep, exactly

PR #601, opened 2016-12-28
bring back 'Member' on usercards

asood123:


frontend/src/components/UserCard.js
```diff
@@ -30,7 +30,7 @@ export default class UserCard extends Component {
 
     const href = `https://opencollective.com/${user.username}`;
 
-    const roleLabel = formatAnchor(user.tier || user.role);
+    const roleLabel = formatAnchor(user.role || user.tier);
```
xdamman: why? What's the rationale here?

asood123: this is what's causing the bug on opencollective.com/tessel. "MEMBER" doesn't  show up because tier is populate as "core-contributor"

xdamman: but I'd argue that the tier should always take precedence. This looks like the wrong fix. Otherwise "sponsor" will never show up (it will show backer which is the role)

asood123: Ah. That stuff is a mess. I don't have an immediate better fix for it. Tiers cause us more small bugs than anything else.

asood123: Why are we overloading these concepts in the first place?

xdamman: ¯\_(ツ)_/¯

xdamman: I'd suggest you put a condition on `if (user.tier === 'core-contributor')`

asood123: uggh we are introducing another string in the mix that's going to need to be tracked on frontend and backend separately? Why do we even need `core contributor`? Isn't it same as `member` everywhere?

xdamman:
member is very confusing because everyone is a member of a collective,
including backers and sponsors and contributors.
core contributors is more appropriate, especially as a "tier". Role should
be ADMIN but let's not get there now. I was just suggesting a quick fix to
solve the issue without creating new issues by swapping precedence of tier
over role.

asood123:
This is a fragile solution that gets us past the current issues on prod. 

It always displays `user.tier` (instead of `user.role`) in both `UserCard` and `HoverCard`.
It converts any tiername into camelCase and looks for that string in `strings.json`. This is to work with existing use cases of RailsGirls and WWCode.
It also has a catchall that if it can't find the tiername in strings.json (like in case of lesbarbares), it returns the string itself.

frontend/src/ui/strings.json
```diff
@@ -298,6 +300,7 @@
     "collectives": "colectivos",
     "comission": "Comisión",
     "contributor": "colaborando",
+    "coreContributors": "Colaborando",
     "coreContributors": "Colaboradores",
```
piamancini: the correct way of saying it is line 304 "coreContributors":"Colaboradores" 
also "coreContributor":"Colaborador"

PR #604, opened 2016-12-29
don't reset expense form on load

asood123:
this fixes submitExpense issue


xdamman:
why?

asood123:
form was being rest after being loaded from `Link` in react-router.

xdamman:
Why was that reset there in the first place?
When you click on "Submit a new expense", does it still reset it properly?

asood123:
yes. not sure why it was there. this code is over 6 months old.

PR #605, opened 2016-12-30
load the collective on componentWillMount

xdamman:


frontend/src/containers/Ledger.js
```diff
@@ -153,34 +153,41 @@ export class Ledger extends Component {
...
-    }
-    Promise.all([
-      fetchUsers(collective.slug),
-      fetchPendingExpenses(collective.slug),
-      fetchTransactions(collective.slug)
-      ])
-    .then(() => scrollToExpense());
+    this.loadCollectiveData()
+      .then((collective) => {
+        this.collective = collective;
```
asood123: I think it's better to use reducers. If the user goes back to the collective page, we'll have to refetch the collective data.

frontend/src/containers/Ledger.js
```diff
@@ -153,34 +153,41 @@ export class Ledger extends Component {
 
   }
 
+  loadCollectiveData() {
+    const { collective, params, fetchProfile } = this.props;
+    if (collective && collective.id) return Promise.resolve(collective);
+    return fetchProfile(params.slug).then(result => {
```
asood123: don't need to return anything here. `fetchProfile` will automatically populate the collective data. See `fetchProfile` in `collective.js`.

frontend/src/containers/Ledger.js
xdamman: I don't understand reducers. How do they work? Should I put the logic of fetching the collective data if not already rendered by the server there?
Isn't it overkill for what we need to do here? "If the user goes back to the collective page": how often would that happen and when it does, how much do we care?

frontend/src/containers/Ledger.js
xdamman: I want to return something because I'm using `(collective)` in the `this.loadCollectiveData().then((collective))`.

frontend/src/containers/Ledger.js
```diff
@@ -221,16 +227,15 @@ export function rejectExp(expenseId) {
...
     fetchTransactions,
     payExpense,
     notify
   } = this.props;
 
-  return payExpense(collective.id, expenseId)
-    .then(() => fetchPendingExpenses(collective.slug))
-    .then(() => fetchTransactions(collective.slug))
+  return payExpense(this.collective.id, expenseId)
+    .then(() => fetchPendingExpenses(this.collective.slug))
```
asood123: Could you add a `notify` here as well? Even better if references the expense data, so you know exactly which one and what was paid. "Expense paid: $61 to Jake Ming (jake@ming.com)" :)

xdamman: Done but without the details. That's more complicated because the api call doesn't return the details of the approved expense.

frontend/src/containers/Ledger.js
asood123: It keeps the data consistent and reusable. It shouldn't be more than few lines change because rest of the work is already done. 

You could do something like this for 
`loadCollectiveData()`: 

```
const { collective, params, fetchProfile } = this.props;
if !(collective && collective.id) {
    fetchProfile(params.slug)
}
```

Then when you are calling `loadCollectiveData()`:

```
this.loadCollectiveData()
 .then(() => Promise.all[...])
 .then(() =>{ switch...})
```

asood123: `collective.js` has the identical example because it loads the same data.

xdamman: I don't understand how this works.
if this.loadCollectiveData doesn't return a promise, how can we be sure that the switch code is run only after fetchProfile returned?

asood123:
closing. this fix is in: https://github.com/OpenCollective/opencollective-website/pull/607

PR #606, opened 2016-12-30
show email if you have permission to see it

asood123:


frontend/src/components/collective/CollectiveExpenseItem.js
```diff
@@ -39,6 +39,13 @@ const CollectiveExpenseItem = ({
 
   let src, srcSet, href, receiptMessage, receiptHelpTip;
 
+  let submittedByName = user && user.name;
+
+  // show email only if the user can view receipt
+  if (canViewReceipt) {
```
xdamman: if canViewReceipt or if canEdit?
Because if we want to allow people to cover the expense in the future, everyone should be able to see it.

asood123: I don't think everyone should be able to see the emails. I think everyone should see the name, yes. I personally wouldn't want to leak my email everywhere on the site.

`canViewReceipt` is different than `canEdit`. `canViewReceipt` includes the submitter of the receipt, in addition to the users that can `canEdit`.

xdamman: I agree. I was just worried that in the future if we allow visitors to cover an expense, they may have the right to canViewReceipt, which would make that email address public. I want to avoid that mistake.

frontend/src/components/collective/CollectiveExpenseItem.js
```diff
@@ -39,6 +39,13 @@ const CollectiveExpenseItem = ({
...
+    submittedByName = user && user.name && user.email ? `${user.name} - ${user.email}` : `${user.email}`;
```
xdamman: I wonder if we should rather show the PayPal email. The email is not confirmed, but the PayPal email is the one where the money will go to. So might double check that value before approving.

frontend/src/components/collective/CollectiveExpenseItem.js
asood123: We don't need to decide that now. This is a one line change, let's get it out. Not setting a precedent here. Besides, the approval process should take care of that. When you `cover`, you trust that since it's approved, it's a legit expense.

frontend/src/components/collective/CollectiveExpenseItem.js
asood123: that's a good idea, i'll update it.

frontend/src/components/collective/CollectiveExpenseItem.js
```diff
@@ -37,13 +37,15 @@ const CollectiveExpenseItem = ({
 
   const updateInProgress = approveInProgress[expense.id] || rejectInProgress[expense.id] || payInProgress[expense.id];
 
+  const userEmail = user && (user.paypalEmail || user.email);
```
xdamman: this will return a boolean

asood123: Works fine for me

xdamman: You are right.
I'm confused 🤔
Time to go to bed here.

Tests were failing. I've restarted them.

asood123: thanks

PR #607, opened 2016-12-30
fix email approve and reject flow and other transitions

asood123:


frontend/src/containers/PublicPage.js
```diff
@@ -18,7 +18,7 @@ export class PublicPage extends Component {
       return <ProfilePage profile={ profile } />
     } else if (this.props.showOldCollectivePage) {
       return <PublicGroup />
-    } else if (this.props.group.slug ) {
+    } else if (this.props.slug ) {
```
xdamman: this is breaking BrusselsTogether.org.
What do you think of replacing this by `if (this.props.slug || this.props.group.slug)`?
Because for BrusselsTogether, the backend hydrates the react app with `this.props.group`

xdamman: Never mind, I ended up using a separate static website for BrusselsTogether.org.

PR #618, opened 2017-01-09
disable submit expense button unless an attachment is included

asood123:


asood123:
This is blocked on fixing e2e test - need to find a way to upload an image through nightwatch tests.

PR #621, opened 2017-01-15
fix login redirect bug user hit on Friday

asood123:
user was confused when /login/:token?next=/login kept taking her back to login page. Now by default, if there is a token and redirect says '/login', it'll take you to homepage.


asood123:
Getting this out, since it's affecting a user in production. @xdamman can you review post-merge?

PR #622, opened 2017-01-17
updated faq

xdamman:
Using the nifty html5 tag `<details>` and `<summary>`


xdamman:
@piamancini @asood123 when you have a chance, you should also add a paragraph or two about #wdydwyd (let's not wait on that to merge this PR though)

PR #624, opened 2017-01-20
default other to one-time donation

xdamman:


asood123:
this fix is not needed, as we already revert back to monthly when you come out of 'Other' dropdown.

PR #627, opened 2017-01-20
Fix donate other

xdamman:
This fixes the bug reported by @piamancini where the `interval` would not go back to the initial interval of the presets after opening the "Other".


asood123:
Looks good. Glad you cleaned up the component.

PR #631, opened 2017-01-25
add confirmation messages when approving, rejecting or paying expense…

asood123:
…s from email

### DIFF
--- a/frontend/src/containers/Ledger.js
+++ b/frontend/src/containers/Ledger.js
@@ -227,6 +227,7 @@ export function approveExp(expenseId) {
 
   return approveExpense(collective.id, expenseId)
     .then(() => fetchPendingExpenses(collective.slug))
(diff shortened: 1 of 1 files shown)


frontend/src/containers/Ledger.js
```diff
@@ -214,6 +214,17 @@ export class Ledger extends Component {
   }
 }
 
+function findExpense(expenseId, expensesObj) {
+  if (expensesObj) {
+    for (let key in expensesObj) { // eslint-disable-line prefer-const
```
xdamman: Can't use a const here? 🤔

It's kind of hacky to do it this way. Ideally we would want to keep those expenses in the store with the key being their id. (I'm not asking to change it, just feedback)

asood123: for loops don't work on second iteration if you use a const there.

yes, ideally expense keyed to their id is best. That's more work than it's worth right now.

PR #644, opened 2017-02-02
Grav customizations

w00fz:
Hi,
this PR includes 3 modifications.

### Grav Customizations

I tweaked some CSS style for Grav's collective to have it more in line with our palette of colors. I hope that is ok!

### Customizations Structure

I reorganized the customizations so that each collective can have their own css file. I moved GulpJS styles into their own CSS and added Grav's to a new separated one. I then changed the `customizations.css` to load these.
I don't think this is the perfect solution. Ideally it would be nice for each collective to be able to  customize settings from the site, such as accent colors, background images and perhaps even light/dark logo variations. 
Although I am just not familiar enough with the code to try tackling this yet. I also noticed that there is a lot of logic in place already for customizations that made me think this is probably already in the list of things to do for you in the future.

### Fixed issue with the ImagePicker

When in editing mode, the first pane of the ImagePicker has a fixed background color of #fafafa. With Grav we have a white and transparent logo which really makes it annoying to look at. Although minor I decided to go ahead and fix this.

The resolution involved just removing the hardcoded background-color. I couldn't see this change affecting anything else in the site beside the logo. Happy to revert back if that's not the case.


asood123:
Nice @w00fz. Thanks for organizing custom css.

PR #654, opened 2017-02-15
css fix for invoice.pdf

xdamman:


xdamman:
merging because a user needs an invoice and this is just a CSS fix.

PR #655, opened 2017-02-17
Assets for email templates in api repo

carlosascari:
For https://github.com/OpenCollective/opencollective-api/pull/802


xdamman:
Can't you use:
https://opencollective.com/static/images/oc-logo-icon.svg 
and 
https://opencollective.com/static/images/oc-logo.svg

carlosascari:
@xdamman 
We can use the `oc-logo-icon.svg` image, although they do differ a bit they are the same image.

`oc-logo.svg` does not exist, I think you meant `logo.svg`, unfortunately the logo's text is a darkish blue color, while white color is needed, so it is not the same image.

I will remove `oc-symbol.svg` and use use the `oc-logo-icon.svg` instead

PR #659, opened 2017-02-23
return a 404 for .well-known

asood123:
Server getting hammered by requests to .well-known. It always hits the database and locks up database, memory, etc.


asood123:
Not needed with nginx changes

PR #662, opened 2017-02-27
Allow to display also inactive/one-time sponsors in embedded list and JSON data

sokra:
Currently the lists doesn't show one-time sponsors, because they are not "active".

This PR fixes this and shows every sponsor regardless of activity.

This aligns the embedded sponsor list with the list on the opencollective website.

### DIFF
--- a/server/src/middlewares.js
+++ b/server/src/middlewares.js
@@ -9,9 +9,10 @@ import _ from 'lodash';
 /**
  * Fetch users by slug
  */
-const fetchActiveUsers = (options = {}) => {
+const fetchUsers = (options = {}) => {
   return (req, res, next) => {
 
+    const requireActive = (typeof options.requireActive === 'boolean') ? options.requireActive : req.query.requireActive !== 'false'; // by default, we skip inactive users
     const filters = {
       tier: req.params.tier,
       exclude: req.query.exclude,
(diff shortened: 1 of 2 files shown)


asood123:
@sokra We made this change on request from several collectives, who only wanted to see the current sponsors showing upon their github page and widget. Maybe it makes more sense to include one-time donators as active users for a period?

sokra:
If "active" is define as doing any donation a one-time sponsor can donate $1/month to be in the list. A small donation makes big difference if you know the system. Are you fine with making this a query option?

sokra:
I changed the PR to keep the original behavior but added a query param `requireActive` which allows to override the behavior.

This should allow to opt-in to disable inactive users in the list.

asood123:
Thanks @sokra

asood123:
Circle CI doesn't run on forks... Let me fork this change so tests run to make sure nothing will break

asood123:
Generated a new PR to get tests to run: https://github.com/OpenCollective/opencollective-website/pull/701

sokra:
@asood123 This worked for a while, but seem to be broken again now.

---

https://opencollective.com/webpack/sponsor/0/avatar?requireActive=true


---

https://opencollective.com/webpack/sponsor/0/avatar?requireActive=false


---

The second one should be Capital One instead like on the opencollective website:

image

PR #665, opened 2017-02-28
support for paypal preapproval on website

asood123:
If there is no valid paypal preapproval:
image

when there is one: 
image


frontend/src/actions/users/confirm_preapproval_key.js
```diff
@@ -0,0 +1,44 @@
+import { post } from '../../lib/api';
+import * as constants from '../../constants/users';
+
+/**
+ * Confirm approval key
+ */
+
```
xdamman: we usually don't add an extra space between the comment of a function and the function.

frontend/src/actions/users/get_approval_key.js
```diff
@@ -0,0 +1,49 @@
...
+import * as constants from '../../constants/users';
+
+/**
+ * Get approval key for paypal
+ */
+
+export default (userid, options={}) => {
+  // Redirect to homepage that will confirm the preapproval key
+  const callback = `${options.baseUrl}/?paypalApprovalStatus=`;
+  const preapprovalTemplate = '${preapprovalKey}';
```
xdamman: Wrong apostrophe. Should use `

xdamman: But why not just `const preapprovalTemplate = preapprovalKey;` ? Need to force it to be a string?

frontend/src/containers/Ledger.js
```diff
@@ -138,6 +150,15 @@ export class Ledger extends Component {
 
         { this.props.route.type === 'expenses' && <div className='Ledger-container padding40 expenses-container'>
             <div className='line1'>unpaid expenses</div>
+
+            {isHost &&
```
xdamman: shouldn't this be `isHost && !paypalCard` ?
I'd rather have that logic outside of the `<PaypalReminder />` component

frontend/src/actions/users/get_approval_key.js
asood123: We don't know the preapproval key at this point in the sequence, paypal fills it out and it needs that format.

frontend/src/containers/Ledger.js
asood123: PaypalReminder needs to decide which message to show and for that it needs the card. It also needs info on details on the card - for example email and eventually a message to expire the card. 

I think it's easier to have a component that's passed a paypalCard and it decides how to display it. We will need to show it other places as well.

xdamman:
ok

PR #666, opened 2017-03-02
logos for emails

xdamman:
For https://github.com/OpenCollective/opencollective-api/pull/802


xdamman:
Merging because it's just assets

PR #668, opened 2017-03-02
put paypal flow behind a flag while we debug with their support

asood123:


asood123:
Not needed right now.

PR #673, opened 2017-03-03
remove anchor from outside of div

asood123:
jsx doesn't link anchor tags outside of div


asood123:
pushing now to get fix out in production, fyi @xdamman

PR #674, opened 2017-03-03
bring our photos back

asood123:


asood123:
merging without waiting for tests, since it's just assets and I'm very late for dinner. :)

PR #677, opened 2017-03-06
Stripe auth

asood123:


asood123:
These changes are in the settings branch.

PR #680, opened 2017-03-08
Settings page with Stripe auth and paypal preapproval

asood123:


frontend/src/lib/utils.js
```diff
@@ -82,4 +82,12 @@ export function scrollToExpense() {
   } else {
    window.scrollTo(0, 0);
   }
+}
+
+export function getUrlHash() {
+    return window.location.hash ? window.location.hash.substr(1).trim().toLowerCase().replace('-', ' ') : '';
```
xdamman: shouldn't we also remove the query params  with `hash.split('?')[0]`?

frontend/src/components/PaypalPreapproval.js
```diff
@@ -2,43 +2,46 @@ import React, { Component, PropTypes } from 'react';
...
       </div>
     );
   }
 }
 
-PaypalReminder.propTypes = {
-  card: PropTypes.object,
+PaypalPreapproval.propTypes = {
+  i18n: PropTypes.object.isRequired,
   onClickConnect: PropTypes.bool.isRequired,
```
asood123: fixed.

frontend/src/components/PaypalPreapproval.js
```diff
@@ -2,43 +2,46 @@ import React, { Component, PropTypes } from 'react';
...
+  card: PropTypes.object,
   onClickInProgress: PropTypes.bool,
-  i18n: PropTypes.object.isRequired,
+  loading: PropTypes.bool,
```
asood123: The component doesn't know when we fetch the paypalCard, so how does it distinguish between no paypal card and loading state on its own?

frontend/src/lib/utils.js
asood123: yea, good idea

frontend/src/components/PaypalPreapproval.js
```diff
@@ -2,43 +2,46 @@ import React, { Component, PropTypes } from 'react';
 
 import ConnectPaypalButton from './ConnectPaypalButton';
 
-class PaypalReminder extends Component {
+class PaypalPreapproval extends Component {
   render() {
     const {
       card,
       onClickConnect,
       onClickInProgress,
```
asood123: You mean onClick?

xdamman: onClick and onChange if needed, standard things.

PR #683, opened 2017-03-09
Awesome site, small oversight

scottdesdev:
Underneath "Our Founding Team" you're losing padding on that paragraph `<p data-reactid=".0.2.2">...</p>` above 900px

What makes the most sense for a quick n' dirty fix here is to simply remove that 600px media query, so you've always got 30px of padding on the ".About-container"

Cheers!

### DIFF
--- a/frontend/src/css/containers/About.css
+++ b/frontend/src/css/containers/About.css
@@ -118,10 +118,8 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


asood123:
Good catch. Thanks for the fix.

PR #685, opened 2017-03-09
new route /:slug/:image(avatar|logo).:format(png|jpg|gif|txt)

xdamman:
that way we can easily get an ASCII version of any collective logo for `npm postinstall` http://github.com/openCollective/opencollective-postinstall


package.json
```diff
@@ -69,6 +69,7 @@
     "express-perf": "1.11.2",
     "express-server-status": "1.0.3",
     "express-session": "1.13.0",
+    "gm": "^1.23.0",
```
asood123: can you remove carat

package.json
```diff
@@ -78,6 +79,7 @@
     "history": "1.17.0",
     "html-pdf": "^2.1.0",
     "image-size": "0.5.0",
+    "image-to-ascii": "^3.0.5",
```
asood123: same thing

PR #689, opened 2017-03-17
Add funds

asood123:
Add Funds is only visible if you are the host of the collective:

screenshot 2017-03-17 15 51 47

By default, all donations come from the host:
screenshot 2017-03-17 15 53 14

You can swap source and then we add a host fee:
screenshot 2017-03-17 15 54 21


frontend/src/components/AddFundsForm.js
```diff
@@ -0,0 +1,164 @@
...
+            <span className='col col-4 AddFundsForm-bb'> Host fees ({collective.hostFeePercent}%): </span>
+            <span className='col col-2 AddFundsForm-host-fees'> {formatCurrency(this.state.hostFees, collective.currency)} </span>
+          </div>
+          <div className='col col-12'>
+            <span className='col col-4'> Net funds for collective:</span>
+            <span className='col col-2 AddFundsForm-amount'> {formatCurrency(this.state.amount - this.state.hostFees, collective.currency)} </span>
+          </div>
+        </div>
+
+        <p className='notice'>When the source of funds is someone other than the host, we deduct the host fee. By clicking below, you agree to set aside these funds in your bank account on behalf of the collective. </p>
```
xdamman: I like this notice!
I'd add in parenthesis the actual amount to be uber clear: "to set aside those funds ($95)"

frontend/src/components/AddFundsForm.js
```diff
@@ -0,0 +1,164 @@
...
+  onAmountChange(updatedAmountText, areFundsFromDefaultUser) {
+    let amount;
+    if (updatedAmountText) {
+      amount = Math.round(100 * updatedAmountText);
+    } else if (updatedAmountText === '') {
+      amount = 0
+    } else {
+      amount = this.state.amount
+    }
+    const addHostFees = !areFundsFromDefaultUser;
```
xdamman: can you rename this variable to `areFundsFromHost`?
I was wondering earlier when reading it "what do you mean by default user?"

frontend/src/css/components/ExpenseForm.css
```diff
@@ -28,7 +28,7 @@
           text-align: center;
           padding-top: 15px;
           font-family: Lato;
-          font-size: 15px;
+          font-size: 14px;
```
xdamman: this may have broken the css of the "add funds" green button

test/unit/containers/Collective.js
```diff
@@ -95,7 +95,7 @@ describe('Collective container', () => {
       expect(payment.stripeToken).to.be.equal(token.id);
       expect(payment.email).to.be.equal(token.email);
       expect(payment.currency).to.be.equal(collective.currency);
-      expect(payment.amount).to.be.equal(10);
+      expect(payment.amount).to.be.equal(1000);
```
xdamman: shouldn't we also add a e2e test for this?
just to make sure that one day we don't have an email from wwcode saying they can't add funds.

frontend/src/components/AddFundsForm.js
asood123: I was trying to keep the component independent of the idea of host. From it's point of view, there is a default funder to show in UI and the user can decide to change that user.

frontend/src/components/AddFundsForm.js
asood123: done

frontend/src/css/components/ExpenseForm.css
asood123: fixed add funds

PR #694, opened 2017-03-24
new route /:slug/donate

xdamman:
e.g. https://opencollective.com/sustainoss/donate


xdamman:
Merging because I have a need for it now for a collective in Brussels. @aseem please review when you can, we can always roll back.

PR #698, opened 2017-04-01
fix for iOS 10.3 and latest Safari on MacSierra

xdamman:


xdamman:
Basically, when `options.headers` is `undefined`, we were calling fetch with `fetch(url, { headers: undefined })` and the native implementation of Safari in iOS 10.3 and Mac Sierra 10.12.4 doesn't like that. It prefers no headers at all or `{ headers: {} }`. The latest build of Safari fixed the issue. But since that bug is part of the stable release pushed by Apple to hundred of millions people, we need this fix.

xdamman:
From: https://developer.apple.com/safari/technology-preview/release-notes/

PR #704, opened 2017-04-10
Fix null twitter handle

xdamman:


frontend/src/containers/ProfilePage.js
```diff
@@ -98,7 +98,7 @@ export class ProfilePage extends Component {
                 placeholder={i18n.getString('defaultWebsite')} />
           <ContentEditable
                 className='ContentEditable-twitterHandle'
-                html={profile.canEditUser ? profile.twitterHandle : `<a href="https://twitter.com/${profile.twitterHandle}" target="_blank">@${profile.twitterHandle}</a>`}
+                html={profile.canEditUser ? profile.twitterHandle : profile.twitterHandle ? `<a href="https://twitter.com/${profile.twitterHandle}" target="_blank">@${profile.twitterHandle}</a>` : ''}
```
asood123: I think multiple conditional operators are confusing to parse. Fine for now, would prefer to avoid them in the future.

PR #716, opened 2017-04-27
allow markdown in tier.description

xdamman:


xdamman:
Merging because Aseem is next to me and reviewed this (after two beers)

PR #723, opened 2017-05-16
add the tags of the host to the new collective

xdamman:
Also added host fee defined by parent host (in `group.settings.hostFeesPercent`, defaults to 5%) and added a new type "coop" as one user was suggesting on Twitter.

For a new host that wants and `/apply`, we need to create a group that has in its settings:

    {
      "HostId": Integer, // User Id of the Host User
      "hostFeesPercent": 5, // optional: percentage for the host (default: 5)
      "superCollectiveTag": "wwcode" // show all collectives that have the same tag on the page
    }

### DIFF
--- a/frontend/src/containers/NewGroup.js
+++ b/frontend/src/containers/NewGroup.js
@@ -17,6 +17,8 @@ import validateSchema from '../actions/form/validate_schema';
 
 import newGroupSchema from '../joi_schemas/newGroup';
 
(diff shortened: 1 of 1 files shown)


xdamman:
I tested locally and it was adding properly the right host. Not sure why it would be different on production.

@piamancini are people creating those collectives using the https://opencollective.com/operationcode/apply url?

PR #728, opened 2017-05-23
OpenCollective customization CSS for codeformiami

ErnieAtLYD:


asood123:
@ErnieAtLYD you'll also need to include this file in: https://github.com/opencollective/opencollective-website/blob/master/frontend/src/css/customization.css

ErnieAtLYD:
@asood123  That was fast. Thank you, sir!

PR #732, opened 2017-06-05
Change reimbursement method

asood123:


asood123:
screenshot 2017-06-05 15 29 09
screenshot 2017-06-05 15 29 15
screenshot 2017-06-05 15 29 29

xdamman:
This is great to finally have this option. Thank you!

I'm not sold on the UX tbh. Why not using a simple dropdown with auto save?
(let's not stop this PR on this but I'm curious)

frontend/src/validators/expense.js
```diff
@@ -28,7 +29,7 @@ const schema = Joi.object().keys({
     .label('IncurredAt'),
   category: Joi.string().required()
     .label('Category'),
-  payoutMethod: Joi.string().valid(pluck(payoutMethods('en'), 'value'))
+  payoutMethod: Joi.string().valid(pluck(payoutMethods(i18n('en')), 'value')) // i18n.getString not used here
```
xdamman: this sounds very hacky. Why having i18n in a validation?
The value of an attribute should be independent of the locale.

asood123:
IF there are two options, i find dropdown to be unnecessary. Dropdown is a UX artifact useful for n choices. iPhone settings is a good example of when to use dropdown vs when to use a switcher.

Also, don't like autosave for something like this. Since money is going to move in a different manner, I'd prefer a confirmation. It's painful to undo if someone sends money to paypal by mistake.

xdamman:
Mmmm... Ok. Maybe add a "instead" at the end to make it more clear it's an alternative choice?

asood123:
changed.

frontend/src/validators/expense.js
xdamman: what about this?

asood123: `payoutMethods()` originally took `lang` and imported i18n on its own. That's worse because we don't always have `lang` available (like in cases where a subcomponent doesn't know about Collective, so can't access lang). I changed it to pass in i18n.

This is an edge case - string doesn't matter, joi only checks for `value`. Refactoring everything to avoid this one hacky line didn't seem worth it.

xdamman: 🤔 Not sure why it's worse given that it defaults to 'en' anyway if there is no lang defined.

(this line is absolutely unreadable... oh well... ¯\_(ツ)_/¯)

PR #735, opened 2017-06-08
allow to accept bitcoin per collective

xdamman:
Stripe transaction fee: 0.8% with $5 max.
See https://stripe.com/bitcoin


piamancini:
woohooo

PR #737, opened 2017-06-09
read host directly from collective

asood123:
Depends on: https://github.com/opencollective/opencollective-api/pull/890


xdamman:
Fixed in https://github.com/opencollective/opencollective-website/pull/742

PR #740, opened 2017-06-12
show link to transactions page

asood123:


xdamman:
We should revisit this. Not a great UX. But will merge to unblock.

PR #745, opened 2017-06-19
only show more transactions on collective page

asood123:


asood123:
Merging to fix wwcode prod issue

PR #746, opened 2017-06-20
fix onclick issue on paid button

asood123:


asood123:
Merging to fix wwcode issue in prod

asood123:
Placing this div inside the same div as actions was breaking the onClick and Hover functions. Clicking on center of button was clicking on the message inside the div for some reason - probably something to do with placements and inlines. Sneaky bug - took me a while to figure out.

PR #753, opened 2017-07-03
Applepay

asood123:


frontend/src/components/Tiers.js
```diff
@@ -107,23 +124,42 @@ export default class Tiers extends Component {
...
-                description={stripeDescription}>
-                  <AsyncButton
-                    color='green'
-                    inProgress={inProgress} >
-                    {button}
-                  </AsyncButton>
-              </StripeCheckout>
-            )}
 
+            {hasStripe && !hasPaypal && (
```
xdamman: Do we really need `!hasPaypal`  here?

frontend/src/components/Tiers.js
```diff
@@ -152,6 +188,37 @@ export default class Tiers extends Component {
       </div>
     );
   }
+
+  callApplePay(tier, payload) {
+    const { collective, onToken } = this.props;
+
+    // setup the payment request
+    const paymentRequest = {
+      countryCode: 'US', // TODO: find out if we need to change this for other countries
```
xdamman: Should this be the country of the platform, or of the Stripe account of the host?
(I'd imagine the latter)

frontend/src/components/Tiers.js
```diff
@@ -152,6 +188,37 @@ export default class Tiers extends Component {
...
+      currencyCode: collective.currency,
+      total: {
+        label: collective.name,
```
xdamman: shouldn't the label be something more explicit (including the `tier.description`)?

xdamman:
Great job on this! Can't wait to push this live!

frontend/src/components/Tiers.js
asood123: don't think we need it either - didn't want to touch the old logic. It's more likely to cause bugs and we are going to retiring this code base anyway.

frontend/src/components/Tiers.js
asood123: unclear from docs. I'll try some variations.

frontend/src/components/Tiers.js
asood123: changed

PR #758, opened 2017-08-02
update list of investors

xdamman:


xdamman:
Merging because only the FAQ has been updated.

PR #771, opened 2017-10-16
change top contributors from 3 to 10 to fix bug for user

asood123:


asood123:
random tests are failing that shouldn't be affected by this change. Merging to unblock user.

PR #773, opened 2017-10-26
creator should be ADMIN of new collective

xdamman:
fixes https://github.com/opencollective/opencollective/issues/563


xdamman:
Keeps on failing with GitHub flow @asood123

PR #777, opened 2017-11-07
Fix apply

asood123:


asood123:
merging, as it's only Github e2e failing on too many attempts.

PR #790, opened 2018-03-09
Use netAmountInCollectiveCurrency instead of amount on user invoices

clarete:


xdamman:
See https://github.com/opencollective/opencollective/issues/949#issuecomment-372383988

I'll look into it

xdamman:
Just pushed a change to this branch to fix https://github.com/opencollective/opencollective/issues/949

PR #800, opened 2018-05-15
deleted tests that are not used anymore

xdamman:


xdamman:
Merging because only the GitHub test doesn't pass (always been finicky)

PR #802, opened 2018-05-17
Further update to the expenses FAQ section

alannallama:
Took out invoicing/address info for the Open Source Collective, because now each host can specify this info in 'expense policies'. The info in the FAQ should apply generally to all hosts.


alannallama:
@xdamman can this be merged?

